PR #53, opened 2020-05-18
ci: fix deps

antonkurenkov:
fixed dependencies


antonkurenkov:
Worklow is fixed, but now I have to check if other ones work correctly (since I was forced to change some dependencies in builder core)

hanxiao:
lgtm!

PR #61, opened 2020-05-19
fix(roberta): put model on cpu

policeme:


hanxiao:
next time, please do not update image just for this level yaml change. this is anti-pattern and not best practice. Instead you should use 

https://github.com/jina-ai/jina-hub#3-build-the-pod-image

image

PR #86, opened 2020-06-15
debug the broken cicd for forked PRs

nan-wang:


nan-wang:
For the debugging purpose only.

PR #87, opened 2020-06-15
feat(encoder): add FaceNet encoder

shivam-raj:
FaceNet Encoder for returning vector embeddings for each face.
Based on FaceNet by : Florian Schroff, Dmitry Kalenichenko, James Philbin ( PyTorch implementation)


shivam-raj:
@nan-wang Hey,
Thanks for your suggestions !
As requested I create this new pull request. 
Can you help me with why commit lint is still failing even after I added subject and type ?

PR #109, opened 2020-07-06
feat(encoder): add distilbert encoder

jayashrimasilamani:


hub/executors/encoders/nlp/transformers-distilbert/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM pytorch/pytorch:1.4-cuda10.1-cudnn7-runtime
+
+RUN pip install transformers --no-cache-dir --compile && \
+    python -c "from transformers import *; x='distilbert-base-cased'; DistilBertModel.from_pretrained(x); DistilBertTokenizer.from_pretrained(x)"
+
+RUN pip install jina
+
+ENTRYPOINT ["jina", "pod", "--yaml-path", "TransformerTorchEncoder"]
```
nan-wang, line 8: This will load  the `TransformerTorchEncoder` with the default model rather than `distilbert` model. Instead of passing a built-in class, you will need specify which model to load in the `yaml`. I suggest to check out the yaml file in the southpark example. [https://github.com/jina-ai/examples/blob/master/southpark-search/pods/encode.yml](https://github.com/jina-ai/examples/blob/master/southpark-search/pods/encode.yml)

PR #136, opened 2020-08-20
[W.I.P] [DO NOT MERGE] feat: add tests for mock ai models for executors

rutujasurve94:
GitHub Issue reference: [Mock AI Models ](https://github.com/jina-ai/jina/issues/824)

- Use input data for testing the *encoding* functionality as an ndarray, eg. B x (Channel x Height x Width)

- Assert the result of encoding against the output data for this encoded data as an ndarray eg. of `B x D` for torchvision.

- This would ensure that the mocking is solely focussed on testing the encoding functionality.

- [ ] Have one unit test per executor model - for all models in image, nlp, etc.
- [ ] Executor test to be dependent on the executor only.
- [ ] Use pytest instead of unittest / mock
-

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,49 @@
+import os
+import shutil
+import sys
+import unittest
+from os.path import dirname
+
+import numpy as np
+
+from jina.drivers.helper import array2pb
+from jina.proto import jina_pb2
+
+
+class JinaTestCase(unittest.TestCase):
+
+    def setUp(self) -> None:
+        self.tmp_files = []
+        os.environ['TEST_WORKDIR'] = os.getcwd()
+
+    def tearDown(self) -> None:
+        for k in self.tmp_files:
+            if os.path.exists(k):
+                if os.path.isfile(k):
+                    os.remove(k)
+                elif os.path.isdir(k):
+                    shutil.rmtree(k, ignore_errors=False, onerror=None)
+
+    def add_tmpfile(self, *path):
+        self.tmp_files.extend(path)
+
+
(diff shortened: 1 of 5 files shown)


JoanFM:
I think that this is not the approach we want to follow in the hub. We do not want to keep the old classes as in the core. Every test should be self contained and not depend on any test base class. Like this we mantain every executor as a small self contained project with its self contained tests.

rutujasurve94:
> I think that this is not the approach we want to follow in the hub. We do not want to keep the old classes as in the core. Every test should be self contained and not depend on any test base class. Like this we mantain every executor as a small self contained project with its self contained tests.

Okay sure, work in progress on de-coupling from the base test classes and making each test independent.
Though few os environment, setup functionality would be common across tests, especially for the same type of executor that could be abstracted out in  ```__init__```.

JoanFM:
> > I think that this is not the approach we want to follow in the hub. We do not want to keep the old classes as in the core. Every test should be self contained and not depend on any test base class. Like this we mantain every executor as a small self contained project with its self contained tests.
> 
> Okay sure, work in progress on de-coupling from the base test classes and making each test independent.
> Though few os environment, setup functionality would be common across tests, especially for the same type of executor that could be abstracted out in `__init__`.

Yes I agree, but since every executor needs to be self contained, I think we do not want to depend on any functionality outside that is not present in the core. I think we trade core reusage for containerization.

nan-wang:
@rutujasurve94 As Joan indicated, we prefer to make the tests of each executor to be self-contained.

PR #139, opened 2020-08-21
Migrate fast ica encoder

YueLiu-jina:
import transformer encoder is base on PR https://github.com/jina-ai/jina/pull/866


encoders/numeric/FastICAEncoder/README.md
```diff
@@ -0,0 +1,6 @@
+# FastICAEncoder
+
+`FastICAEncoder` encodes data from an ndarray in size `B x T` into an ndarray in size `B x D`.
+
+note:
```
JoanFM, line 5: that is wrong under current level, that was before, now u can pass a pickled file from which to do encoding

encoders/numeric/FastICAEncoder/tests/test_fasticaencoder.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+import pickle
+
+
+from .. import FastICAEncoder
+from encoders.numeric.test import TestClassNumeric
```
JoanFM, line 6: we do not want to have these kind of dependencies, every test needs to be self contained, so every test will have to implement its tests (test_encode, test_load ....)

encoders/numeric/test.py
```diff
@@ -0,0 +1,111 @@
+import os
```
JoanFM, line 1: This needs to be removed, every executor needs have its own self contained tests and no dependency from other files in the tree

JoanFM (changes requested):

JoanFM:
This is a missconception on how we have to tests executors (see comments also added in #136)

nan-wang:
@YueLiu-jina I'm looking into this PR

hanxiao (approved):

hanxiao (approved):

encoders/numeric/FastICAEncoder/requirements.txt
```diff
@@ -0,0 +1 @@
+sklearn
```
JamesTang-jinaai, line 1: should be `scikit-learn`

PR #147, opened 2020-08-24
Fix sptagindexer

JoanFM:
This is one step to fix the docker build, but test is still failing because the code does not do what it should

### DIFF
--- a/indexers/vector/NmsLibIndexer/Dockerfile
+++ b/indexers/vector/NmsLibIndexer/Dockerfile
@@ -4,6 +4,7 @@ FROM jinaai/jina
 COPY . /workspace
 WORKDIR /workspace
 
+RUN apt-get update && apt-get install --no-install-recommends -y gcc libboost-all-dev libgsl0-dev libeigen3-dev
 # install the third-party requirements
 RUN pip install -r requirements.txt
 
--- a/indexers/vector/NmsLibIndexer/Dockerfile
+++ b/indexers/vector/NmsLibIndexer/Dockerfile
@@ -4,7 +4,6 @@ FROM jinaai/jina
 COPY . /workspace
 WORKDIR /workspace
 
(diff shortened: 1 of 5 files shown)


indexers/vector/SptagIndexer/__init__.py
```diff
@@ -44,11 +44,15 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
         _index.SetBuildParam("NumberOfThreads", str(self.num_threads))
         _index.SetBuildParam("DistCalcMethod", self.space)
 
-        if _index.Build(vecs, vecs.shape[0]):
+        if _index.Build(vecs.astype('float32'), vecs.shape[0]):
```
JoanFM, line 47: In their example, they Build, save to a new directory and load from the directory? Is it needed here? And also they have a different mode where they use a client to a remote indexer

indexers/vector/NmsLibIndexer/Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM jinaai/jina
 COPY . /workspace
 WORKDIR /workspace
 
+RUN apt-get update && apt-get install --no-install-recommends -y gcc libboost-all-dev libgsl0-dev libeigen3-dev
```
JoanFM, line 7: Remove this as not part of the PR?

indexers/vector/SptagIndexer/__init__.py
nan-wang: I'm following this tutorial, https://github.com/microsoft/SPTAG/blob/master/docs/Tutorial.ipynb

JoanFM: Yes, this is the one I am talking about, but maybe the `Save` is just for persistence and not necessary. If the tests pass it is good

nan-wang: Yes, `Save` is  for saving the index for later usages. It is not required by the `Search`.

PR #163, opened 2020-08-24
fix: remove jina from faiss requirements

JoanFM:
Problem with ruamel-yaml when trying to reinstall `jina`. Is 0.15.89 required for `jina` to really work? is `0.15.87` enough?


nan-wang:
> Is 0.15.89 required for jina to really work? is 0.15.87 enough?

What's the issue here?

nan-wang (approved):
LGTM👍

JoanFM:
> > Is 0.15.89 required for jina to really work? is 0.15.87 enough?
> 
> What's the issue here?

The issue is that faiss cpu reinstalls ruamel to be 0.15.87 and whem jina is again in requirements it tries to install 0.15.89 but an exception occurs that it cannot safelly uninstall it because it is a distutils package

PR #175, opened 2020-08-25
refactor: migrate ONNX Encoder

rutujasurve94:
Migrating Onnx Encoder


encoders/image/OnnxImageEncoder/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    pip install -r requirements.txt
+
+# RUN pip install -r requirements.txt
+COPY . /
+
+# for downloading pre-trained model
+RUN bash download.sh
+
+# for testing the image
+RUN pip install pytest && pytest -v
```
JoanFM, line 18: I would remove `RUN bash download.sh` and I would do:

`RUN download.sh && pip install pytest && pytest -v -s && rm -rf mobilenetv2-1.0`

so that we do not upload an image with a model in it, since at running time one might use a different one or can map a `volume`

rutujasurve94: Yes sure, updated the Dockerfile with the changes

nan-wang: Does the `download.sh` work for you? In my case, it ends up with an empty file.

nan-wang:
It's better to switch to `git-lfs` for downloading the models. https://github.com/onnx/models#usage---git-lfs-

PR #258, opened 2020-08-27
feat: add laser encoder

bhavsarpratik:
Added LASER encoder in nlp


hanxiao (approved):

JoanFM:
Laser was already used in our multilingual example; we should make sure to change it to use this version in the hub.

PR #295, opened 2020-08-27
chore: bump BigTransferEncoder version to 0.0.3

hanxiao:


nan-wang:
The failure of the unit test can *NOT* be reproduced locally.

nan-wang:
The `.yml` config file saved inside the docker is broken. The following part is missing,

```
with:
  channel_axis: 1
  model_path: pretrained
```

Example config file saved during the test
```
!BigTransferEncoder
metas:
  name: BigTransferEncoder-f8a9e621
  replica_id: 0
  replica_workspace: .//None-{root.metas.replica_id}
  separated_workspace: false
requests:
  on:
    ControlRequest:
    - !ControlReqDriver
      with:
        priority: 0
    IndexRequest:
    - &id001 !EncodeDriver
      with:
        apply_order: post
        depth_range:
        - 0
        - 1
        executor: null
        method: encode
        priority: 0
        traverse_on:
        - chunks
    SearchRequest:
    - *id001
    TrainRequest:
    - *id001
```

nan-wang:
When running the following codes locally and in docker containers, the `encoder._init_kwargs_dict` differes. Inside docker, this variable is empty while locally the `model_path` is shown. This leads to the different results when dumping to yaml.

```python
from __init__ import BigTransferEncoder
encoder = BigTransferEncoder(model_path='./pretrained')
print(encoder._init_kwargs_dict)
```

nan-wang:
This issue is due to the usage of `ExecutorType` in jina.

Usually, a customer-defined executor `Foo` has `ExectutorType` as the meta class. 

```
class Foo(BaseExcutor):
    ...

class BaseExecutor(metaclass=ExecutorType):
        ...

class ExecutorType(type):
    def __new__():
        ...

```

In the `__new__()` method, `Foo` is registered via `register_class()`. During this process, the arguments of the `__init__` method are added to `_init_kwargs_dict` variable via `store_init_kwargs()`. The `_init_kwargs_dict` variable is later used for saving the configs.

**The problem** is that we are using `git submodule` to maintain the executors under the `jina/hub` folder. All the classes under `jina/hub` are registered when importing `jina`. When 'Foo' that is already in `jina/hub`, the problem happens. In this case, if we try to import a local version of `Foo` by `from . import Foo`, the arguments of the `__init__()` will **NOT** be added to `_init_kwargs_dict` because `Foo` is already registered and the `store_init_kwargs()` will **NOT** be called.

Take an example,
```
from jina.hub.encoders.image.BigTransferEncoder import BigTransferEncoder
encoder = BigTransferEncoder(model_path='./pretrained')
print(encoder._init_kwargs_dict)   # result: {'model_path': './pretrained', 'channel_axis': 1}
```

```
from . import BigTransferEncoder
encoder = BigTransferEncoder(model_path='./pretrained')
print(encoder._init_kwargs_dict)   # result: {}
```

hanxiao:
I don't get it why this face the problem while others don't?

JamesTang-jinaai:
While debugging the test_save_and_load method and running the code below
`_output = self.model(self._get_input(data.astype(np.float32)))` 
I encountered the problem which tells me that `variable was uninitialized`. That reason is that `pickle` didn't save `tensorflow variables` successfully. In my test case, I just need to use `encoder.save()`. So here are two solutions:
1. work around to dump `tensorflow variables` by `pickle` with some parameters, which seems impossible;
2. separate tensorflow variables and other objects into 2 methods in base Class. You may want to check [this](https://stackoverflow.com/questions/38000180/save-tensorflow-model-to-file)

Here is the log:
```Log
2020-08-30 20:36:55.087455: W tensorflow/core/common_runtime/base_collective_executor.cc:217] BaseCollectiveExecutor::StartAbort Failed precondition: Error while reading resource variable resnet/block4/unit02/b/group_norm/gamma_109626 from Container: localhost. This could mean that the variable was uninitialized. Not found: Resource localhost/resnet/block4/unit02/b/group_norm/gamma_109626/N10tensorflow3VarE does not exist.
	 [[{{node StatefulPartitionedCall/resnet/block4/unit02/b/group_norm/Reshape_2/ReadVariableOp}}]]
Traceback (most recent call last):
  File "/opt/pycharm-2020.2/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "/software/jina/jina/executors/decorators.py", line 158, in arg_wrapper
    return func(self, data, *args, **kwargs)
  File "/software/jina/jina/executors/decorators.py", line 58, in arg_wrapper
    r = func(self, *args, **kwargs)
  File "/software/jina-hub/encoders/image/BigTransferEncoder/__init__.py", line 72, in encode
    _output = self.model(self._get_input(data.astype(np.float32)))
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/eager/function.py", line 1551, in __call__
    return self._call_impl(args, kwargs)
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/eager/function.py", line 1591, in _call_impl
    return self._call_flat(args, self.captured_inputs, cancellation_manager)
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/saved_model/load.py", line 99, in _call_flat
    cancellation_manager)
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/eager/function.py", line 1692, in _call_flat
    ctx, args, cancellation_manager=cancellation_manager))
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/eager/function.py", line 545, in call
    ctx=ctx)
  File "/home/jamestang/.pyenv/versions/3.7.6/lib/python3.7/site-packages/tensorflow_core/python/eager/execute.py", line 67, in quick_execute
    six.raise_from(core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.FailedPreconditionError:  Error while reading resource variable resnet/block4/unit02/b/group_norm/gamma_109626 from Container: localhost. This could mean that the variable was uninitialized. Not found: Resource localhost/resnet/block4/unit02/b/group_norm/gamma_109626/N10tensorflow3VarE does not exist.
	 [[{{node StatefulPartitionedCall/resnet/block4/unit02/b/group_norm/Reshape_2/ReadVariableOp}}]] [Op:__inference_signature_wrapper_62023]
Function call stack:
signature_wrapper
```
Here are the variables in `encoder`, and you can notice that the `tensorflow variables` are in the list
encoder_trainable_variables

Here are the variables in `encoder_loaded`, and you can notice that the `tensorflow variables` are _NOT_ in the list
encoder_loaded_trainable_variables

PR #321, opened 2020-08-27
fix: laser and scan import

hanxiao:


hanxiao:
@theUnkownName @bhavsarpratik please find the rational behind inline-import
https://github.com/jina-ai/jina/issues/373#issuecomment-623483847

theUnkownName (approved):

Issue #332, opened 2020-09-03
misspelling in TransformerTFEncoder test case


JamesTang-jinaai:
Parameter pooling_strategy was misspelled to polling_strategy. Because default value was set, it just leads to duplicated test cases.

┆Issue is synchronized with this [Jira Task](https://jinaai.atlassian.net/browse/JINAHUB-57) by [Unito](https://www.unito.io/learn-more)

PR #335, opened 2020-09-03
chore: changed misspelling

JamesTang-jinaai:
Resolved: #332


JoanFM:
I think is good, but are we sure all the tests passed? The commit message should not be `chore` since it actually affects `testing`.

hanxiao (approved):

PR #374, opened 2020-09-03
chore: bump BigTransferEncoder version to 0.0.6

hanxiao:


JoanFM:
Depends on local model path and if set to None will fail at post init

maximilianwerk (approved):
Nice cleanup. LGTM

PR #400, opened 2020-09-06
refactor: dataclass in transformer encoder

bhavsarpratik:
Refactoring as per jina-ai/jina#910
Making `TransformerTorchEncoder` itself dataclass is problematic due to inheritance, args and kwargs.
This PR is an experimental suggestion.


JoanFM:
Hey @bhavsarpratik,

I think is a cool feature, but I personally do not see a great benefit that compensated the boilerplate of integrating it everywhere, plus since it is not such a known feature I guess many contributors would not easily comply with it.

Let's see what other contributors may think.

bhavsarpratik:
Hi @JoanFM, sure I am also thinking about the pros/cons of using dataclass. This is an experimental suggestion to write shorter code. I hope this doesn't break documentation or any other thing.

hanxiao:
aha, the way @bhavsarpratik is using `dataclass` is actually pretty smart 👍 . it makes me think...

hanxiao (approved):

hanxiao:
I will merge it as it does simplify the code by leveraging the native Py37 feature. I have to think how can this be applied to all executors without adding extra step in the learning curve.

PR #403, opened 2020-09-07
PDF text extractor

theUnkownName:
Extracts text from PDF.
All images will be discarded


crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    :class:`PDFTextExtractor` Extracts text from PDF.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+    def craft(self, uri: str, *args, **kwargs):
+        import PyPDF2
```
JoanFM, line 18: move the import to `post_init`

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+    """
+    :class:`PDFTextExtractor` Extracts text from PDF.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+    def craft(self, uri: str, *args, **kwargs):
```
JoanFM, line 17: what if the input is not a local file but it comes as `bytes`?

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+        import PyPDF2
+        pdf_obj = open(uri, 'rb')
```
JoanFM, line 19: this file needs to be also properly closed

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+        pdf_reader = PyPDF2.PdfFileReader(pdf_obj)
+        page_obj = pdf_reader.getPage(0)
```
JoanFM, line 21: why do we just take first page? The library offers the `read(self, stream)` interface that I think takes all text

crafters/nlp/PDFExtractor/tests/test_pdfextractor.py
```diff
@@ -0,0 +1,25 @@
...
+                              "In posuere auctor pellentesque. Donec cursus ipsum ut odio dictum ultricies. "
+                              "Suspendisse volutpat accumsan imperdiet. Pellentesque sodales pharetra egestas. S"
+                              "ed tincidunt nec eros vel mollis. Etiam blandit leo vitae mattis bibendum. "
+                              "Nunc rutrum hendrerit ligula. ")
+    canvas.save()
+
+def test_read_pdf():
+    crafter = PDFTextExtractor()
+    text = crafter.craft('cats_are_awesome.pdf')
+    print(text)
```
JoanFM, line 24: u can have the text in a static variable and check the extracted value

JoanFM (changes requested):

crafters/nlp/PDFExtractor/__init__.py
theUnkownName: Right! I missed this

crafters/nlp/PDFExtractor/tests/test_pdfextractor.py
```diff
@@ -0,0 +1,25 @@
+from .. import PDFTextExtractor
+
+def create_pdf():
+    from reportlab.pdfgen.canvas import Canvas
```
JoanFM, line 4: if you use the page interface, make sure you are creating a pdf with multiple pages

theUnkownName: Yeah the example one has 2 pages with one image to test that

crafters/nlp/PDFExtractor/tests/test_pdfextractor.py
```diff
@@ -1,25 +1,18 @@
...
     crafter = PDFTextExtractor()
-    text = crafter.craft('cats_are_awesome.pdf')
+    text = crafter.craft(uri='cats_are_awesome.pdf', buffer=None)
     print(text)
 
+
+def test_io_buffer():
+    import requests
+    url = 'http://ceur-ws.org/Vol-2410/paper35.pdf'
+    response = requests.get(url)
```
JoanFM, line 14: Download or create a complicated PDF (open source one if downloaded), But do not add requests in our test.

1. Won't work if no connection available.
2. What if they remove this pdf from the site? We would not have a clue what's wrong

JoanFM (changes requested):

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -14,11 +14,10 @@ def __init__(self, *args, **kwargs):
 
     def craft(self, uri: str, buffer: bytes, *args, **kwargs):
         import PyPDF2
-        import io
 
         text = ""
         if buffer:
-            pdf_obj = io.BytesIO(buffer)
+            pdf_obj = buffer
```
JoanFM, line 20: buffer object has no `close`. PdfFileReader accepts a `stream` that is what `io.BytesIO` provides.

crafters/nlp/PDFExtractor/tests/test_pdfextractor.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+
+def test_io_uri():
+    crafter = PDFTextExtractor()
+    text = crafter.craft(uri='cats_are_awesome.pdf', buffer=None)
+    print(text)
+
+
+def test_io_buffer():
```
JoanFM, line 11:
```suggestion
def test_io_buffer():
    with open('paper35.pdf', 'rb') as fp:
        input_bytes = fp.read()
        
     crafter = PDFTextExtractor()
     text = crafter.craft(uri=None, buffer=input_bytes)
     assert text == expected_text
```

JoanFM (changes requested):

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -12,23 +12,29 @@ class PDFTextExtractor(BaseCrafter):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    def craft(self, uri: str, buffer: bytes, *args, **kwargs):
+    def post_init(self):
         import PyPDF2
+        import io
 
+    def craft(self, uri: str, buffer: bytes, *args, **kwargs):
+        import PyPDF2
```
JoanFM, line 20: remove these imports then

crafters/nlp/PDFExtractor/__init__.py
```diff
@@ -12,23 +12,29 @@ class PDFTextExtractor(BaseCrafter):
...
-            page = pdf_reader.getPage(i)
-            text += page.extractText()
+        try:
+            pdf_reader = PyPDF2.PdfFileReader(pdf_obj)
+            count = pdf_reader.numPages
+            for i in range(count):
+                page = pdf_reader.getPage(i)
+                text += page.extractText()
+        except IOError:
+            print("Could not open file")
```
JoanFM, line 37: do not handle `exception` here, it is handled by `jina` at different levels depending on configurations.

crafters/nlp/PDFExtractor/tests/test_pdfextractor.py
```diff
@@ -1,15 +1,20 @@
 from .. import PDFTextExtractor
 
 
+expected_text = "A cat poem\nI love cats, I love every kind of cat,\nI just wanna hug all of them, but I can't," \
```
JoanFM, line 4: Perfect, how does it then behave with pdfs that are in chinese?

theUnkownName: Looks like it's not encoding those characters, they are being ignored at the moment

PR #406, opened 2020-09-09
Feat levenshtein distance ranker

maximilianwerk:
Adding a Levenshtein distance ranker to the hub. That might not be to useful in practice, but rather gives an example how to use the `Match2DocRanker`.

### DIFF
--- a/rankers/LevenshteinRanker/tests/test_levenshteinranker.py
+++ b/rankers/LevenshteinRanker/tests/test_levenshteinranker.py
@@ -20,7 +20,7 @@ def test_levenshteinranker():
         copy.deepcopy(match_meta)
     )
 
-    assert (new_scores == np.array([(1, 0), (2, -3)], dtype=np.float64)).all()
+    assert np.testing.assert_array_equal(new_scores, [(1, 0), (2, -3)])
     # Guarantee no side-effects happen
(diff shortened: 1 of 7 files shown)


maximilianwerk:
The build will only work, after a new jina-ai/jina docker image is build, since it uses the `Match2DocRanker` that was just introduced in the jina-core master.

JoanFM (dismissed):

rankers/LevenshteinRanker/tests/test_levenshteinranker.py
```diff
@@ -0,0 +1,27 @@
...
+    match_meta_json = json.dumps(match_meta, sort_keys=True)
+
+    ranker = LevenshteinRanker()
+    new_scores = ranker.score(
+        copy.deepcopy(query_meta),
+        copy.deepcopy(old_match_scores),
+        copy.deepcopy(match_meta)
+    )
+
+    assert (new_scores == np.array([(1, 0), (2, -3)], dtype=np.float64)).all()
```
JoanFM, line 23: I think you can use `np.testing.assertEqual` or something like this, I think it is more clear like that

JoanFM (dismissed):

JoanFM (approved):

PR #407, opened 2020-09-09
feat: added reverse ranker

maximilianwerk:
A ranker that reverses the previous ranking order. As in #406: The build will only work, after a new jina-ai/jina docker image is build, since it uses the Match2DocRanker that was just introduced in the jina-core master.

### DIFF
--- a/rankers/ReverseRanker/Dockerfile
+++ b/rankers/ReverseRanker/Dockerfile
@@ -1,0 +1,13 @@
+FROM jinaai/jina
+
+# setup the workspace
+COPY . /workspace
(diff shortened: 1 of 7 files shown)


hanxiao:
isn't this `ReverseQL` driver?

maximilianwerk:
Fair enough. So the `ReverseQL` is intended to be used at any step after the retrieval?

hanxiao:
you can chain the driver in `requests.on:`

```
        - !Chunk2DocRankerDriver {}
        - !ReverseQL {}
```

PR #418, opened 2020-09-22
feat: add batch_size as meta param

JoanFM:


indexers/vector/NmsLibIndexer/__init__.py
```diff
@@ -45,13 +49,11 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
         _index.createIndex({'post': 2}, print_progress=self.print_progress)
         return _index
 
-    @batching(batch_size=512)
+    @batching()
```
hanxiao, line 52:
```suggestion
    @batching
```

no need

PR #420, opened 2020-09-22
perf: annoy reserves max(idex) + 1 memory, reverse may be more efficient

JoanFM:
**Changes introduced**
According to annoy documentation.

```
a.add_item(i, v) adds item i (any nonnegative integer) with vector v. Note that it will allocate memory for max(i)+1 items.
```
so, adding items in reverse order should save calls to `malloc` on Annoy end.


JoanFM:
I close since it seems unnecessary early optimization

PR #421, opened 2020-09-23
fix: fix model_name default parameter for ImageTorchEncoder

### LINKED ISSUE #419: ImageTorchEncoder always loads mobilenet_v2
The class ImageTorchEncoder sets 'mobilenet_v2' as default CNN, but the ordering of the variables should be inverted from: 

https://github.com/jina-ai/jina-hub/blob/b2f41f7bfe978e9ef8b9688bf3d089114de3734e/encoders/image/ImageTorchEncoder/__init__.py#L47

to 

`self.model_name = model_name or 'mobilenet_v2'`

for proper overriding of the default.

fsal:
Fix #419

### DIFF
--- a/encoders/image/ImageTorchEncoder/__init__.py
+++ b/encoders/image/ImageTorchEncoder/__init__.py
@@ -17,8 +17,8 @@ class ImageTorchEncoder(BaseTorchEncoder):
 
     def __init__(
         self,
-        model_name: str = "mobilenet_v2",
-        pool_strategy: str = "mean",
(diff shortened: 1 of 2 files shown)


JoanFM:
Hey!

Thank you very much for the PR, in jina we use `single quotes` as our style. Could you please revert this style reverted changes?

JoanFM:
also, you will need to update the `version` in the `manifest.yml` for it to be built and tested

JoanFM (approved):

PR #423, opened 2020-09-23
Zarr Indexer

deepankarm:
Indexer based on Zarr 

- Inherits `NumpyIndexer`
- `create_handler` & `add_handler` are based on opening a `Zarr Group`, inside which we write a `Zarr Array`
- `query_handler` reads the `Zarr Array`
- Avoid using `[:]` (which gives access to core np array), as that decreases the performance. 
- Memory usage is equivalent to `np.memmap`

Refs :
[Zarr Docs](https://zarr.readthedocs.io/en/stable/index.html)
[Zarr vs Python mmap](https://pythonspeed.com/articles/mmap-vs-zarr-hdf5/)

### DIFF
--- a/indexers/vector/ZarrIndexer/__init__.py
+++ b/indexers/vector/ZarrIndexer/__init__.py
@@ -49,10 +49,8 @@ class ZarrIndexer(NumpyIndexer):
     def get_query_handler(self) -> Optional['zarr.core.Array']:
         if not (path.exists(self.index_abspath) or self.num_dim or self.dtype):
             return
-
-        if hasattr(self, 'write_handler') and 'default' in self.write_handler.array_keys():
-            return zarr.open(store=f'{self.index_abspath}/default', mode='r',
-                             shape=(self._size, self.num_dim), chunks=True)
+        return zarr.open(store=f'{self.index_abspath}/default', mode='r',
+                         shape=(self._size, self.num_dim), chunks=True)
     
     def query_by_id(self, ids: Union[List[int], 'np.ndarray'], *args, **kwargs) -> 'np.ndarray':
         int_ids = [self.ext2int_id[j] for j in ids]
--- a/indexers/vector/ZarrIndexer/__init__.py
+++ b/indexers/vector/ZarrIndexer/__init__.py
@@ -1,0 +1,63 @@
+from os import path
+from typing import List, Union, Optional
+
+import zarr
+import numpy as np
+
+from jina.executors.indexers.vector import NumpyIndexer
+from jina.helper import cached_property
+
+
+class ZarrIndexer(NumpyIndexer):
+    """Indexing based on Zarr arrays
+    
+    For more information about Zarr, please check
+    https://zarr.readthedocs.io/en/stable/index.html
+    
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        
+    def get_add_handler(self) -> zarr.hierarchy.Group:
+        """Open an existing zarr group file for adding new vectors
+
+        :return: a zarr group file
+        """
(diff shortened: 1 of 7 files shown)


indexers/vector/ZarrIndexer/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    def get_create_handler(self) -> zarr.hierarchy.Group:
+        """Create a zarr group file for adding new vectors
+
+        :return: a zarr group file
+        """
+        return zarr.open(store=self.index_abspath, mode='w')
+    
+    def add(self, keys: 'np.ndarray', vectors: 'np.ndarray', *args, **kwargs) -> None:
+        self._validate_key_vector_shapes(keys, vectors)
+        if 'default' in self.write_handler.array_keys():
```
JoanFM, line 37: what does this do?

indexers/vector/ZarrIndexer/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_create_handler(self) -> zarr.hierarchy.Group:
+        """Create a zarr group file for adding new vectors
+
+        :return: a zarr group file
+        """
+        return zarr.open(store=self.index_abspath, mode='w')
+    
+    def add(self, keys: 'np.ndarray', vectors: 'np.ndarray', *args, **kwargs) -> None:
+        self._validate_key_vector_shapes(keys, vectors)
```
JoanFM, line 36: Is this validated in every addition for other indexers?

indexers/vector/ZarrIndexer/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    
+    @property
+    def query_handler(self):
+        return self.get_query_handler()
+    
+    def get_query_handler(self) -> Optional['zarr.core.Array']:
+        if not (path.exists(self.index_abspath) or self.num_dim or self.dtype):
+            return
+
+        if hasattr(self, 'write_handler') and 'default' in self.write_handler.array_keys():
```
JoanFM, line 53: it feels weird to depend on `write_handler` when querying, why is this option needed?

indexers/vector/ZarrIndexer/__init__.py
deepankarm: We are creating a group first, since we cannot know the array size, and without knowing the array size. we cannot create a `zarr.core.Array`. And inside that group, we are creating an array called `default`. This is done in the `add` method using `write_handler`. Hence in `query_handler`, we are just verifying that.

indexers/vector/ZarrIndexer/__init__.py
deepankarm: It is done in the same way in `NumpyIndexer`

indexers/vector/ZarrIndexer/__init__.py
JoanFM: `query_handler` is normally used without the `write_handler` initialized I think, at least I think is how most of the times it behaves, since we are splitting index and querying time

indexers/vector/ZarrIndexer/__init__.py
deepankarm: This is just to check if the `default` array is present in the group that was created during `write_handler` creations. Instead, we can open the `zarr Group` first, then check for `default` array. That would be 2 `zarr.open` calls. This seems better.

indexers/vector/ZarrIndexer/__init__.py
deepankarm: Looking at other Indexers, I have removed this check. Directly reading the array now,

JoanFM (approved):

PR #424, opened 2020-09-25
Test advanced indexers

JoanFM:


JoanFM:
Some of these tests fail, I think they do not offer guarantees that the index they return is has any correspondance with what is stored and expected in `self.int2ext_id`

indexers/vector/NmsLibIndexer/tests/test_nmslibindexer.py
```diff
@@ -1,69 +1,109 @@
...
+        save_abspath = indexer.save_abspath
+
+    with BaseIndexer.load(save_abspath) as indexer:
+        assert isinstance(indexer, NmsLibIndexer)
+        idx, dist = indexer.query(queries, top_k=1)
+        np.testing.assert_equal(idx, np.array(
+            [[10000], [11000], [12000], [13000], [14000], [15000], [16000], [17000], [18000], [19000]]))
+        assert idx.shape == dist.shape
+        assert idx.shape == (10, 1)
+        np.testing.assert_equal(indexer.query_by_id([10000, 15000]), vectors[[0, 5000]])
```
nan-wang, line 109:
```suggestion
    vectors = np.random.uniform(low=5.0, high=10.0, size=(10000, 1024)).astype('float32')

    queries = vectors[:10, :]

    keys = np.arange(10000, 20000).reshape(-1, 1)

    with NmsLibIndexer(index_filename='nmslib.test.gz', num_threads=4, metas=metas) as indexer:
        indexer.add(keys, vectors)
        indexer.save()
        assert os.path.exists(indexer.index_abspath)
        save_abspath = indexer.save_abspath

    with BaseIndexer.load(save_abspath) as indexer:
        assert isinstance(indexer, NmsLibIndexer)
        idx, dist = indexer.query(queries, top_k=1)
        np.testing.assert_equal(idx, np.arange(10000, 10010).reshape(-1, 1))
        assert idx.shape == dist.shape
        assert idx.shape == (10, 1)
        np.testing.assert_equal(indexer.query_by_id([10000, 15000]), vectors[[0, 5000]])
```

PR #427, opened 2020-09-26
Tika extractor

fhaase2:
https://tika.apache.org/

> The Apache Tika™ toolkit detects and extracts metadata and text from over a thousand different file types (such as PPT, XLS, and PDF). All of these file types can be parsed through a single interface, making Tika useful for search engine indexing, content analysis, translation, and much more. You can find the latest release on the download page. Please see the Getting Started page for more information on how to start using Tika.


crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+            try:
+                response = requests.get(TIKA_URL)
+            except ConnectionError:
+                pass
+
+    def __init__(self, *args, **kwargs):
+        self.start_tika()
+        super().__init__(*args, **kwargs)
+
+    def craft(self, uri: str, *args, **kwargs):
```
JoanFM, line 37: Is there a chance for it to accept bytes? And can it return images as well as text, if so it would be cool to have it as a Segmenter

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+TIKA_URL = 'http://0.0.0.0:9998'
+
+
+class TikaExtractor(BaseCrafter):
+    """
+    :class:`TikaExtractor` Extracts text from files.
+    """
+
+    def start_tika(self):
```
JoanFM, line 18: How will this process be finished? Should there be some handling in the overriden close method?

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+        result = parser.from_file(filename=uri,
+                                  serverEndpoint=TIKA_URL,
+                                  service='all',
+                                  xmlContent=False,
+                                  headers=headers,
+                                  requestOptions=request_options)
+        if 'status' in result:
+            if result['status'] == 200:
+                extract['text'] = result['content']
+                extract['metadata'] = result['metadata']
```
JoanFM, line 58: There is no metadata in our proto message and metainfo will be soon deprecated, we want to rncourage the usage of tags, you can see the document proto definition under proto/jina.proto

JoanFM (changes requested):

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,60 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import os
+
+import requests
+from requests.exceptions import ConnectionError
+from jina.executors.crafters import BaseCrafter
+
+TIKA_URL = 'http://0.0.0.0:9998'
```
JoanFM, line 10: I guess this is the url to connect to the jvm underlying right? Isnt there some py4j connection with Tika?

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class TikaExtractor(BaseCrafter):
+    """
+    :class:`TikaExtractor` Extracts text from files.
+    """
+
+    def start_tika(self):
+        from subprocess import Popen
+        tika_version = os.getenv('TIKA_VERSION')
```
JoanFM, line 20: It seems a little odd to have this depend on an env variable?

crafters/nlp/TikaExtractor/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM logicalspark/docker-tikaserver
+
+WORKDIR /
+
+# get python
+RUN apt-get update && apt-get upgrade -y && apt-get clean
```
JoanFM, line 6: Can u put all these layers together so that not so many partial images are created? This helps to obtain a smaller docker image

crafters/nlp/TikaExtractor/__init__.py
fhaase2: Done.

crafters/nlp/TikaExtractor/__init__.py
fhaase2: Killed the process now in the overriden `close` method.

crafters/nlp/TikaExtractor/__init__.py
fhaase2: Its the connection to the REST api provided by Tika. Yes you are right, might be possible to directly connect to the jvm

crafters/nlp/TikaExtractor/__init__.py
fhaase2: The env variable is set by the upstream docker image

crafters/nlp/TikaExtractor/Dockerfile
fhaase2: Done

crafters/nlp/TikaExtractor/__init__.py
theUnkownName: Hi cool! just wanted to ask the same, can it work with images too?

fhaase2: Yes, Tika will apply tesseract ocr internally AFAIK.

theUnkownName: Nice! I asked cause I tried it too but wasn't working very well for me in the tests, but cool you made it work!
But then if it returns both I think it should be on the Segmenter, since Crafters are supposed to work with 1-1 documents

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+                                      headers=headers,
+                                      requestOptions=request_options)
+        else:
+            raise ValueError('No value found in "buffer" and "uri"')
+
+        if 'status' in result:
+            if result['status'] == 200:
+                text = result['content']
+
+        return text
```
JoanFM, line 75: The return type should be a `dict` containing `text` in the key text. This will be then processed by the driver and interpreted as a `jina_pb2.Document`

crafters/nlp/TikaExtractor/__init__.py
JoanFM: Yes, is there a possibility that `tika` returns more than one `result` in output? Like an output for text and some other output for images? or how does it behave on that?

fhaase2: Tika would return text anyway. It would not return the detected images. So the input could be images, pdfs, docx, ... but output would be always text. In this case it should be a crafter?

JoanFM: Yes, in this case it should be a crafter yes! Then in any downstream segmenter u could split all the text as u want.

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    def post_init(self):
+        super().post_init()
+        self.start_tika()
+
+    def start_tika(self):
+        from subprocess import Popen
+        tika_version = os.getenv('TIKA_VERSION')
+        self.tika_process = Popen(
+            ['java', '-jar', f'/tika-server-{tika_version}.jar', '-h', '0.0.0.0'],
+            stdout=open('stdout.log', 'w'),
```
JoanFM, line 30: Avoid creating files better? can't u pipe it to stdout and stderr ?

crafters/nlp/TikaExtractor/__init__.py
JoanFM: u can see other crafters to see how they return a dictionary

JoanFM (changes requested):

theUnkownName:
Hi @fhaase2 , I see you said it only returns text, but as you mentioned it also works with images, would it be possible to also return the images? or what happens with a PDF with text and images for example?

fhaase2:
Hi @theUnkownName, its only returns text. 
**Input**: images, pdf, docx, emails, ... 
**Output**: extracted text

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Retry until tika server is started
+        for _ in range(10):
+            try:
+                time.sleep(1)
+                requests.get(TIKA_URL)
+                return
+            except ConnectionError:
+                pass
+        sys.exit(0)
```
JoanFM, line 45: Better raise Exception

JoanFM (changes requested):

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def close(self):
```
JoanFM, line 47: please call `super().close()` as well

JoanFM (changes requested):

crafters/nlp/TikaExtractor/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        raise TimeoutError('Timeout when waiting for tika to start')
+
+    def close(self):
+        super().close()
+        self.tika_process.kill()
+
+    def craft(self, uri: str, buffer: bytes, *args, **kwargs):
+        from tika import parser
+        headers = {
+            'X-Tika-PDFOcrStrategy': os.getenv('TIKA_OCR_STRATEGY', 'ocr_only'),
```
JoanFM, line 54: all these env variables, it is best if they are passed as parameters to the Executor I think

JoanFM (changes requested):

crafters/nlp/TikaExtractor/__init__.py
fhaase2: Okay sure, do you have an example executor where we pass parameters?

JoanFM: I mean in the constructor of the Executor, many executors take parameters with some default values

JoanFM (changes requested):
Even tika version could be passed as param?

fhaase2:
> Even tika version could be passed as param?

I am not sure about this, because it is set by the upstream docker image. When they publish a new version they will bump the version

JoanFM (approved):

PR #430, opened 2020-10-01
feat: keras uses tree traversal

maximilianwerk:
Adapting `ImageKerasEncoder` for the new tree traversal already used in the encoders.

### DIFF
--- a/encoders/image/ImageKerasEncoder/config.yml
+++ b/encoders/image/ImageKerasEncoder/config.yml
@@ -10,8 +10,7 @@ requests:
     [IndexRequest, SearchRequest]:
       - !EncodeDriver
         with:
-          granularity_range: [1, 1]
-          recur_on: ['chunks', ]
+          traversal_paths: ['c']
 metas:
   py_modules:
     - __init__.py
--- a/encoders/image/ImageKerasEncoder/config.yml
+++ b/encoders/image/ImageKerasEncoder/config.yml
@@ -3,14 +3,6 @@ with:
   model_name: MobileNetV2
   pool_strategy: avg
   channel_axis: -1
-requests:
-  on:
-    ControlRequest:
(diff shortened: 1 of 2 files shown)


JoanFM:
u need to bump manifest version for it to really activate tests and publish!

encoders/image/ImageKerasEncoder/config.yml
```diff
@@ -10,8 +10,7 @@ requests:
     [IndexRequest, SearchRequest]:
       - !EncodeDriver
         with:
-          granularity_range: [1, 1]
```
JoanFM, line 13: I think Drivers should be removed by default, is this the only encoder with a Driver attached different than the default ones in the config.yml?

maximilianwerk: Yes it is.

JoanFM: So let's remove it, I think we should keep default config, any user can later adapt it, but for coherence as well I think Drivers should not be changed by default.

maximilianwerk: No, what I meant is, that it is the only executor in the hub, that explicitly does not need the default, yet (which is `['c', 'r']`.

JoanFM: I do not see why this should behave differently than ImageTorchEncoder?

rutujasurve94 (dismissed):
LGTM

encoders/image/ImageKerasEncoder/config.yml
JoanFM: And why shouldn't it be used to encode the root?

JoanFM (changes requested):

JoanFM (approved):

PR #431, opened 2020-10-04
feat: add compression vae executor

umbertogriffo:
# Description
This pull request provides to the user an additional tool to encode data from a ndarray.
**CompressionVAE** is a fast, easy to use, dimensionality reduction tool based on the idea of [Variational Autoencoders](https://www.jeremyjordan.me/variational-autoencoders/). We can consider it an alternative to **t-SNE** and **UMAP**.

* **Pros**:
  * **CompressionVAE** is faster than either **t-SNE** or **UMAP**.
  * **CompressionVAE** learns a deterministic and reversible mapping from data to embedding space.
  * It can be integrated into live systems that can process previously unseen examples without re-training.
  * It can in principle scale to arbitrarily large datasets.
  * **VAEs** have a very strong and well studied theoretical foundation.

* **Cons**:
  * As almost any deep learning tool, it needs a lot of data to be trained. **t-SNE** and **UMAP** work better on small datasets.

# Future work 
The current implementation could be adapted to be used with images, text and videos applying convolutional or recurrent encoders/decoders.

# References
* [Variational autoencoders](https://www.jeremyjordan.me/variational-autoencoders/)
* [CompressionVAE — A Powerful and Versatile Alternative to t-SNE and UMAP](https://towardsdatascience.com/compressionvae-a-powerful-and-versatile-alternative-to-t-sne-and-umap-5c50898b8696)
* [CompressionVAE Full code and documentation](https://github.com/maxfrenzel/CompressionVAE)

### DIFF
--- a/encoders/numeric/CompressionVaeEncoder/__init__.py
+++ b/encoders/numeric/CompressionVaeEncoder/__init__.py
@@ -16,13 +16,20 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
     Full code and documentation can be found here: https://github.com/maxfrenzel/CompressionVAE..
     """
 
-    def __init__(self, model_path, train_valid_split=0.99, output_dim=16,
+    def __init__(self, model_path='temp', X=None, train_valid_split=0.99, output_dim=16,
                  iaf_flow_length=10, cells_encoder=[512, 256, 128],
                  initializer='lecun_normal', batch_size=32, batch_size_test=32,
                  feature_normalization=False, tb_logging=False,
                  *args, **kwargs):
         """
         :param model_path: specifies the path to the model, and also acts as the model name.
+        :param X : array, shape (n_samples, n_features)
+            Training data for the VAE.
+            Alternatively, X can be the path to a root-directory containing npy files (potentially nested), each
+            representing a single feature vector. This allows for handling of datasets that are too large to fit
+            in memory.
+            Can be None (default) only if a model with this name has previously been trained. Otherwise None will
+            raise an exception.
         :param train_valid_split: controls the random splitting into train and test data. Here 99% of X is used
         for training, and only 1% is reserved for validation.
         :param dim_latent: specifies the dimensionality of the latent space.
@@ -38,6 +45,7 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
         :param tb_logging: determines whether the model writes summaries for TensorBoard during the training process.
         """
         super().__init__(*args, **kwargs)
+        self.X = X
         self.train_valid_split = train_valid_split
         self.output_dim = output_dim
         self.iaf_flow_length = iaf_flow_length
@@ -61,7 +69,7 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
         :return: a `B x D` numpy ``ndarray``
         """
         from cvae import cvae
-        model = cvae.CompressionVAE(data,
+        model = cvae.CompressionVAE(X=self.X,
                                     train_valid_split=self.train_valid_split,
                                     dim_latent=self.output_dim,
                                     iaf_flow_length=self.iaf_flow_length,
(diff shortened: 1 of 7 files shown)


encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+                                    dim_latent=self.output_dim,
+                                    iaf_flow_length=self.iaf_flow_length,
+                                    cells_encoder=self.cells_encoder,
+                                    initializer=self.initializer,
+                                    batch_size=self.batch_size,
+                                    batch_size_test=self.batch_size_test,
+                                    logdir=self.model_path,
+                                    feature_normalization=self.feature_normalization,
+                                    tb_logging=self.tb_logging)
+
```
JoanFM, line 75: Is there a way to use a pretrained model? With this implementation a model will be trained for every batch.

JoanFM (changes requested):

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: Yes, you are right. I think we have the same problem of the [TSNEEncoder](https://github.com/jina-ai/jina-hub/blob/master/encoders/numeric/TSNEEncoder/__init__.py) where the `fit_transform` method fits each batch into an embedded space and return that transformed output. 

Theoretically, we can use pre-trained models, but they don't exist. To generate them, we need to train this architecture with datasets that could make sense in this context.

What if we use [as_train_method](https://github.com/jina-ai/jina/blob/df7a7f494be7683a8377fdcf3bb8036ed7238e0a/jina/executors/decorators.py#L34) and [require_train](https://github.com/jina-ai/jina/blob/df7a7f494be7683a8377fdcf3bb8036ed7238e0a/jina/executors/decorators.py#L71) decorator functions to run the training one time before calling the encode function? or should I use a [PipelineEncoder](https://github.com/jina-ai/jina/blob/df7a7f494be7683a8377fdcf3bb8036ed7238e0a/jina/executors/encoders/__init__.py#L68)?

JoanFM: The encoder for now must assume that it loads a pretrained model. It should not be training for now before loading. See as example CustomImageTorchEncoder, where a pretrained path is passed. TSNEEncoder is a spefial case because only fit transform is provided but because the algorithm is as such by design.

encoders/numeric/CompressionVaeEncoder/tests/test_compressionvaeencoder.py
```diff
@@ -41,23 +41,37 @@ def target_output_dim():
     return target_output_dim
 
 
-def test_encoding_results(test_data, target_output_dim):
-    expected_batch_size = test_data.shape[0]
+@pytest.fixture(scope="function")
+def get_encoder(test_data, target_output_dim):
+    from cvae import cvae
+    model_path = tempfile.NamedTemporaryFile().name
```
JoanFM, line 47: can you use `tmpdir` pytest fixture?

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: Ok got it. I've updated the code, so now we can load a pretrained model. 

I'm wondering if in the future will make sense to create a specific TransformEncoder, that now is sklearn based, for other frameworks like Tensorflow, Keras and PyTorch.

JoanFM (changes requested):

JoanFM (approved):

encoders/numeric/CompressionVaeEncoder/tests/test_compressionvaeencoder.py
umbertogriffo: Yes, good catch.

JoanFM:
Thank you very much for your contribution!

PR #432, opened 2020-10-05
fix: crafter image resizer target_size list

bhavsarpratik:


JoanFM:
To activate proper tests passing u must bump the version!

crafters/image/ImageResizer/tests/test_imageresizer.py
```diff
@@ -10,7 +10,15 @@ def test_resize():
     img_width = 20
     img_height = 17
     output_dim = 71
+    
+    # Test for int target_size
     crafter = ImageResizer(target_size=output_dim)
-    img_array = create_random_img_array(img_width, img_height)
+    img_array = create_random_img_array(img_height, img_width)
```
JoanFM, line 16: I think the point is to change the order in the function arguments and not in the call.

JoanFM: no my bad!

JoanFM (approved):

crafters/image/ImageResizer/tests/test_imageresizer.py
```diff
@@ -10,7 +10,15 @@ def test_resize():
...
     crafted_doc = crafter.craft(img_array)
     assert min(crafted_doc['blob'].shape[:-1]) == output_dim
+
+    # Test for tuple/list target_size
+    crafter = ImageResizer(target_size=[img_height, img_width])
+    img_array = create_random_img_array(img_height, img_width)
```
JoanFM, line 22: better to have a different input size than the resize?

JoanFM (changes requested):

crafters/image/ImageResizer/tests/test_imageresizer.py
bhavsarpratik: Couldn't get you. The target_size=int will resize any of the smaller size. I changed this to match the conventions of the `random array` function and tuple of target_dim which expects `(height, width)`

bhavsarpratik:
I guess all is ready

crafters/image/ImageResizer/tests/test_imageresizer.py
JoanFM: yes yes, my bad, I always get my head messy with num_rows, num_cols, widths and heights!

JoanFM (approved):

PR #433, opened 2020-10-05
feat: proper close

JoanFM:
Since the addition of FluentdHandler, in the base executor there is a call to properly close the executor `logger`. So all the executors that override the close behavior should ensure that they call the parent close.

Besides this, PaddleHubEncoder was missing some dependencies to work properly


maximilianwerk (approved):
I am not exactly sure, why all of the tests got that much slimmer, but I like it.

PR #442, opened 2020-10-08
fix: parameter overwrite enabled again

maximilianwerk:
This is a revert from some time ago. Alternative solution would be:
```
    super().init(self, *args, **kwargs)
    Transformer(self, *args, **kwargs)
```

Multiclass inheritance is not always working as one would expect.

### DIFF
--- a/encoders/nlp/TransformerTFEncoder/__init__.py
+++ b/encoders/nlp/TransformerTFEncoder/__init__.py
@@ -1,7 +1,6 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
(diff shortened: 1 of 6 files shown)


maximilianwerk:
That one fixed the quality issue I was talking about.

JoanFM (approved):

JoanFM (changes requested):
Please bump versions

JoanFM:
> Please bump versions

plus it would be nice to add a test overriding params again

maximilianwerk:
I removed some tests for `TransformerTFEncoder`, which anyhow never ran correctly before and never succeeded.

JoanFM (approved):

PR #444, opened 2020-10-09
fix: load model in post_init

JoanFM:
Hey @umbertogriffo,

I realized a problem at least with `cvae 0.0.3`.

- First, from jina perspective , we can't instantiate a model at every batch, therefore the model loading should be done at `post_init` time.

- Second, when initializing model, it removes the `logdir`, so I do not understand how it is working for the current version (because I also see they raise an exception about X not being of the right type (the check of X being None is commented out).

Could you point what is the direction to have to load a pretrained model from a path (currently jina executor should not care for now about `training` params).


umbertogriffo:
@JoanFM I've opened [448](https://github.com/jina-ai/jina-hub/pull/448) where I fixed the problem.

PR #447, opened 2020-10-10
fix: added tuple support

bhavsarpratik:
_resize_short accepts only tuple. Fixes jina-ai/examples#235

### DIFF
--- a/crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
+++ b/crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
@@ -15,7 +15,7 @@ def test_transform_results():
     assert crafted_doc["blob"].shape == (224, 224, 3)
 
     img_size = (217, 200)
-    target_size = (224, 224)
+    target_size = (100, 224)
     crafter = ImageNormalizer(target_size=target_size)
     img_array = create_random_img_array(*img_size)
     crafted_doc = crafter.craft(img_array)
--- a/crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
+++ b/crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
@@ -13,3 +13,10 @@ def test_transform_results():
     img_array = create_random_img_array(img_size, img_size)
     crafted_doc = crafter.craft(img_array)
     assert crafted_doc["blob"].shape == (224, 224, 3)
+
+    img_size = (217, 200)
+    target_size = (224, 224)
+    crafter = ImageNormalizer(target_size=target_size)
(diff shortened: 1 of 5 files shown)


JoanFM:
The PR title is missleading because it also adds a new parameter.

I honestly do not see why would someone would use a scale by different than 255. It is an ImageNormalizer not an arrayNormalizer ... I personally would not apply this change. 

On the other hand, u must bump manifest version to run the tests.

JoanFM (changes requested):
I would not add the scale by param

bhavsarpratik:
I am looking to have the flexibility to scale or not scale. I can also keep `scale: bool = True` instead of scale_by but scale_by seemed more flexible. I felt forced scaling can be a problem if I am using normalisation many times in the flow. Currently for object example, torchobject again scales up by 255 and so its not a problem if I apply normalisation on object images.

Ahh I keep forgetting the version bump.

JoanFM:
> I am looking to have the flexibility to scale or not scale. I can also keep `scale: bool = True` instead of scale_by but scale_by seemed more flexible. I felt forced scaling can be a problem if I am using normalisation many times in the flow. Currently for object example, torchobject again scales up by 255 and so its not a problem if I apply normalisation on object images.
> 
> Ahh I keep forgetting the version bump.

Can u describe in detail a Flow where u are seeing the need for this change? If needed, I would go for scale_by, but lets have a detail description of a case where it is needed.

bhavsarpratik:
I think it will become more clear when I push the object example. It doesn't require the argument for now. So let me remove it and push this.

JoanFM:
Can you add a couple of tests where tuple input is accepted?

crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
```diff
@@ -13,3 +13,10 @@ def test_transform_results():
     img_array = create_random_img_array(img_size, img_size)
     crafted_doc = crafter.craft(img_array)
     assert crafted_doc["blob"].shape == (224, 224, 3)
+
+    img_size = (217, 200)
+    target_size = (224, 224)
```
JoanFM, line 18: And a test to have different width and height? That would be the point actually

JoanFM (changes requested):

crafters/image/ImageNormalizer/tests/test_imagenormalizer.py
bhavsarpratik: Giving one side only resizes short while giving tuple sizes both. The input has different h, w and so it has to change both.
I will push your change though.

JoanFM (approved):

Issue #445, opened 2020-10-09
CompressionVAEEncoder instantiates model for every batch


JoanFM:
Linked to draft #444 

I realized a problem at least with cvae 0.0.3.

First, from jina perspective , we can't instantiate a model at every batch, therefore the model loading should be done at post_init time.

Second, when initializing model, it removes the logdir, so I do not understand how it is working for the current version (because I also see they raise an exception about X not being of the right type (the check of X being None is commented out).

Could you point what is the direction to have to load a pretrained model from a path (currently jina executor should not care for now about training params).

JoanFM:
@umbertogriffo

JoanFM:
Closed by #448

PR #372, opened 2020-09-03
chore: bump CustomImageTorchEncoder version to 0.0.5

hanxiao:

PR #448, opened 2020-10-11
fix: load model in post_init

umbertogriffo:
Related to the issue [445](https://github.com/jina-ai/jina-hub/issues/445) and the draft [444](https://github.com/jina-ai/jina-hub/issues/445).

The model is loaded at `post_init` time just providing the model path. Instead of initializing the model using the `CompressionVAE` object, we can directly initialize the model using the `VAEModel` object configured with the params.json and the checkpoint of the pre-trained model.


encoders/numeric/CompressionVaeEncoder/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM jinaai/jina
+
+# setup the workspace
+COPY . /workspace
+WORKDIR /workspace
+
+RUN python3 -m pip install --upgrade pip
```
JoanFM, line 7: Why id this needed?

umbertogriffo: It's not mandatory, I added that cmd just to avoid the `WARNING: You are using pip version 20.0.2; however, version 20.2.3 is available.`

JoanFM: If want to do this pleade merge it with another command so that we do not create an extra docker layer

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,68 +17,60 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
+
+            self.embeddings = net.embed(self.data_feature_placeholder)
 
-    @batching
+            self.sess = tf.Session(config=config)
+            init = tf.global_variables_initializer()
+            self.sess.run(init)
+
+            # Saver for loading checkpoints of the model.
+            saver = tf.train.Saver(var_list=tf.trainable_variables(), max_to_keep=2)
```
JoanFM: Wjat isthis max to keep param?

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,68 +17,60 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
+            self.data_feature_placeholder = tf.placeholder_with_default(
+                input=tf.zeros([64, param['dim_feature']], dtype=tf.float32),
+                shape=[None, param['dim_feature']])
+
+            self.embeddings = net.embed(self.data_feature_placeholder)
 
-    @batching
+            self.sess = tf.Session(config=config)
+            init = tf.global_variables_initializer()
+            self.sess.run(init)
```
JoanFM: Who is responsible for closing this session? Should a close be added to close it?

JoanFM (changes requested):

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,68 +17,60 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
+
+            # Saver for loading checkpoints of the model.
+            saver = tf.train.Saver(var_list=tf.trainable_variables(), max_to_keep=2)
+            cvae.load(saver, self.sess, self.model_path)
+
+            self.to_device()
+
+    @batching(batch_size=64)
```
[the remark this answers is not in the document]
umbertogriffo: Ok, nice.

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: It's the maximum number of model checkpoints to keep. It's used in training but here is useless, I'm going to drop it.

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -62,12 +62,12 @@ def post_init(self):
...
 
                 # Saver for loading checkpoints of the model.
-                saver = tf.train.Saver(var_list=tf.trainable_variables(), max_to_keep=2)
+                saver = tf.train.Saver(var_list=tf.trainable_variables())
                 cvae.load(saver, self.sess, self.model_path)
 
                 self.to_device()
 
-    @batching(batch_size=64)
+    @batching()
```
JoanFM, line 70: Simply `batching` without () works

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,68 +18,63 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
+        import cvae.lib.model_iaf as model
+        import tensorflow as tf
+
+        params_path = os.path.join(self.model_path, 'params.json')
+
+        config = tf.ConfigProto(log_device_placement=False)
+        config.gpu_options.allow_growth = True
+
+        with tf.Graph().as_default():
+            if os.path.exists(params_path):
```
JoanFM, line 41: Raise exception otherwise

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,68 +18,63 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
-                                    initializer=self.initializer,
-                                    batch_size=self.batch_size,
-                                    batch_size_test=self.batch_size_test,
-                                    logdir=self.model_path,
-                                    feature_normalization=self.feature_normalization,
-                                    tb_logging=self.tb_logging)
-        return model.embed(data)
+        import tensorflow as tf
+
+        return self.sess.run([self.embeddings],
```
JoanFM, line 79: Just cueious, what are the return values of session run? Why is the first index the one we want in return?

JoanFM (changes requested):

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: We can't add a close there because in that case we are going to run the encode function on an already closed session. So the post_init can't be responsible for that. I think the session should be closed by the encode function. 
I'm going to use `with self.sess as sess:` because `with` uses a context manager which already handles the entry and exit from its block of code. So we don't need to call close().

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: right

encoders/numeric/CompressionVaeEncoder/__init__.py
JoanFM: Yes I agree post init shouldnnot be respobsible for sure, but neither should encode since it may run more than once.

The point is about overriding `close` method of Executor. And even better but not needed to inherit ExitStack and enter context of session to have it context managed during executor lifetime, but that may not necessary.

I would just override close and ensure tjat we call parent class close and close the session.

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -67,14 +67,14 @@ def post_init(self):
...
     def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
         """
         :param data: a `B x T` numpy ``ndarray``, `B` is the size of the batch
         :return: a `B x D` numpy ``ndarray``
         """
         import tensorflow as tf
-
-        return self.sess.run([self.embeddings],
-                             feed_dict={self.data_feature_placeholder: data})[0]
+        with self.sess as sess:
```
JoanFM, line 78: Would that contrxg manager revuild the session? Would it be more effective to have only one during executor lifetime?

umbertogriffo: Should I use @cached_property to cache it?

JoanFM: No, I do not think so, my point is one only needs to close it once

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: Now I got your point. I agree and I think it's the proper way to handle it.

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: It returns a ndarray with shape (1 x batch size x number of features of embedding) where 1 means one batch.

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -68,16 +76,10 @@ def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
...
-                                    feature_normalization=self.feature_normalization,
-                                    tb_logging=self.tb_logging)
-        return model.embed(data)
+        import tensorflow as tf
+
+        return self.sess.run([self.embeddings],
+                             feed_dict={self.data_feature_placeholder: data})[0]
+
+    def close(self) -> None:
+        self.sess.close()
```
JoanFM, line 85: Can u add a super.close() call to ensure anything added on base classes' close is also executed?

JoanFM (changes requested):

JoanFM (approved):

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: Sure

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: I raised the exception, but now we are always getting that exception because we don't have an existent pre-trained model. How did you manage it for the CustomImageTorchEncoder since the CI calls it with an empty model_path?

JoanFM: I did pickled a modem as a preprocess of the test and provide that model path in the constructor of the encoder. I think such a logic can also be applied here?

umbertogriffo: Yes, the problem is not the unit tests where I did what you said and they succeed, but when the container runs again and no model_path is provided.

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,50 +18,56 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
-        self.batch_size = batch_size
-        self.batch_size_test = batch_size_test
         self.model_path = model_path
-        self.feature_normalization = feature_normalization
-        self.tb_logging = tb_logging
 
     def post_init(self):
-        super().post_init()
-        self.to_device()
+        from cvae import cvae
```
JoanFM, line 31: Call super().post_init()

JoanFM (changes requested):

JoanFM:
Ok I see, yes we have a problem that we still do not know how to go about having default values in config.yml for executors depending on pretrained model paths. Lets keep it as it is, I think now the PR is correct and see how we figure that out

JoanFM:
Hey, waiting for the patch to pass, but since merging https://github.com/jina-ai/jina/pull/1065, now it would be good to go if at `post_init` the exception `PretrainedModelFileDoesNotExist` is raised when not pretrained model can be found.

JoanFM:
> Hey, waiting for the patch to pass, but since merging [jina-ai/jina#1065](https://github.com/jina-ai/jina/pull/1065), now it would be good to go if at `post_init` the exception `PretrainedModelFileDoesNotExist` is raised when not pretrained model can be found.

It is already available, you can see #372 as an example

umbertogriffo:
@JoanFM It worked! Thank you very much for the patch and for your time!

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -33,42 +34,45 @@ def post_init(self):
...
-                self.sess.run(init)
-
-                # Saver for loading checkpoints of the model.
-                saver = tf.train.Saver(var_list=tf.trainable_variables())
-                cvae.load(saver, self.sess, self.model_path)
-
-                self.to_device()
-            else:
-                raise FileNotFoundError(f'The {params_path} doesn\'t exist')
+        if self.model_path and os.path.exists(self.model_path):
```
JoanFM, line 37: It seems that the new logic and the FileNotFound exception should have the same logic, is there a way to unify it? Maybe checking model path and params path before session run?

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -16,50 +19,60 @@ class CompressionVaeEncoder(BaseNumericEncoder, BaseTFEncoder):
...
 
     def post_init(self):
         super().post_init()
-        self.to_device()
+        from cvae import cvae
+        import cvae.lib.model_iaf as model
+        import tensorflow as tf
+
+        if self.model_path and os.path.exists(self.model_path):
+            params_path = os.path.join(self.model_path, 'params.json')
```
JoanFM, line 38: You can skip the check on `model_path` and do it for `params_path`:

```python
params_path = os.path.join(self.model_path, 'params.json') if self.model_path else None

if params_path and os.path.exists(self.model_path):
...
else:
raise PretrainedModelFileDoesNotExist()
```
```

**Review by JoanFM — CHANGES_REQUESTED** [review]

File: encoders/numeric/CompressionVaeEncoder/__init__.py (hunk 12 above)
Comment (umbertogriffo) [reply encoders/numeric/CompressionVaeEncoder/__init__.py:38]: I think it would be better assigning to params_path an existent model path or None, and check if the JSON file is really contained in the folder afterwards:

```.py
params_path = os.path.join(self.model_path, 'params.json') if self.model_path and os.path.exists(
            self.model_path) else None

if self.params_path and os.path.exists(params_path):
...
else:
raise PretrainedModelFileDoesNotExist()
```

because otherwise, we don't know if the folder contains that file. And if the JSON for any reason in missing we are going to get an Exception when we are loading it.

JoanFM: perfect! It seems good!

encoders/numeric/CompressionVaeEncoder/__init__.py
```diff
@@ -34,45 +34,45 @@ def post_init(self):
         import cvae.lib.model_iaf as model
         import tensorflow as tf
 
-        if self.model_path and os.path.exists(self.model_path):
-            params_path = os.path.join(self.model_path, 'params.json')
+        params_path = os.path.join(self.model_path, 'params.json') if self.model_path and os.path.exists(
+            self.model_path) else None
+
+        if self.params_path and os.path.exists(params_path):
```
JoanFM, line 40: no `self` here

encoders/numeric/CompressionVaeEncoder/__init__.py
umbertogriffo: Nice! I'm going to push that change.

JoanFM (approved):

JoanFM:
Hey @umbertogriffo ,

unfortunately there is a problem with the commit message:

Bad commit message, please refer to:
- [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
- [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)
#553 :bookmark: Commit Message Guideline for the First Time Contributor
jina-ai/jina | Jun 21st | Added by GitHub
CONTRIBUTING.md
jina-ai/jina | Added by GitHub

umbertogriffo:
@JoanFM yes I've put a dot at the end of the sentence.

JoanFM:
> @JoanFM yes I've put a dot at the end of the sentence.

Yes, it is a little picky ;)

JoanFM:
Thanks for the effort and the patience @umbertogriffo !!

umbertogriffo:
@JoanFM Glad to contribute! And thanks for your support!

PR #451, opened 2020-10-11
ci: add cla.yml

JoanFM:


.github/workflows/cla.yml
```diff
@@ -0,0 +1,26 @@
...
+        # Alpha Release
+        uses: hanxiao/github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          path-to-signatures: '.github/signatures/v1/cla.json'
+          path-To-cladocument: 'https://github.com/jina-ai/jina/blob/master/CLA.md'
+          # branch should not be protected
+          branch: 'cla'
+          whitelist: hanxiao,nan-wang,jina-bot,BingHo1013,alexcg1,imsergiy,YueLiu-jina,JoanFM,lusloher,bwanglzu
```
maximilianwerk, line 24: Should there be all the engineers or just specific ones? If all, can we add the missing ones? Or would it be possible to just add the jina-team and automagically everyone of the team is inside?

JoanFM: @hanxiao is it possible that this `whitelist` is not needed anymore since the team passes automatically? Should it only be used to grant special treatment to some external contributors?

JoanFM: @maximilianwerk I did not even read it, I just copied what we have in the core ;)

maximilianwerk: Reading is only for the reviewers...

JoanFM: @nan-wang do you know what is the expected value here?

hanxiao: i guess yes, but i'm not the author of this github action https://github.com/cla-assistant/github-action

nan-wang: We can skip the whitelist for the time being because all the team members have committed. I didn't figure out a way to add a group of users on the white list. 

BTW, I've just noticed that the `whitelist` is renamed to `allow list` recently. 
https://github.com/cla-assistant/github-action/blob/fad4db8e82099233366bd88c0ef5c20adaf12a12/CHANGELOG.md#v200-alpha-2020-08-22

PR #457, opened 2020-10-15
test: sptag advanced indexer testing

JoanFM:
Split of the PR #424 for different indexers

There is a test that I skipped because it fails but I think is because we failed to find the proper configuration of parameters to make it work. SPTAG has not really good documentation


bwanglzu (approved):
lgtm!

deepankarm (approved):
Looks good

PR #461, opened 2020-10-26
feat: setup ndcg evaluator

bwanglzu:
Add `NDCGEvaluator`, associated documents and test cases.


evaluators/rank/NdcgEvaluator/requirements.txt
```diff
@@ -0,0 +1 @@
+jina
```
nan-wang, line 1:
```suggestion
jina>=0.7.0
```

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -14,9 +15,29 @@ def metric(self):
...
+
+    def _compute_dcg(self, gains):
+        """Compute discounted cumulative gain."""
         ret = 0.0
-        pass
+        for score, position in zip(gains[1:], range(2, len(gains) + 1)):
+            ret += score/log(position, 2)
+        return gains[0] + ret
+
+    def _compute_idcg(self, gains):
```
JoanFM, line 40: I think these 2 can be made @static right?

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -0,0 +1,22 @@
...
+
+@pytest.fixture(scope='function')
+def evaluator():
+    return NDCGEvaluator(eval_at=3)
+
+@pytest.fixture(scope='function')
+def desired():
+    return [.8, .4, .1]
+
+@pytest.mark.parametrize('actual, expected',[
```
JoanFM, line 13: make sure to test edge cases, where some of them are empty and so on...

JoanFM (changes requested):

JoanFM:
Please have a small `README.md` in the same way other ones have.

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: @JoanFM why was that?

JoanFM: these 2 functions do not relly in any attribute of `self`?

JoanFM:
> Please have a small `README.md` in the same way other ones have.

Okey @bwanglzu , right now in the `README.md` you need to specify that the `actual` expected values are `relevancy` scores and not `ids` as per `Precision` and `Recall`.

JoanFM:
and you can put an example on how we could use it with `RankEvaluateDriver` and an `id_tag` of `score` in `tags` of the `protobuf` document.

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: @JoanFM I get you point :) mostly we do this since it's not rely on any attribute of `self` and use it outside the class. Such as:
```suggestion
    @staticmethod
    def do_something(param):
            # do something, mostly utility purpose, without rely on self but the logic belongs to this class
            pass
          
```

And we can call it with `NDCGEvaluator.do_something(param = param)`

bwanglzu: since this function is called internally, i personally think it's useless to add the decorator.

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+        return f'nDCG@{self.eval_at}'
+
+    def evaluate(self, actual: Sequence[Any], desired: Sequence[Any], *args, **kwargs) -> float:
+        """"
+        :param actual: the scores predicted by the search system.
+        :param desired: the expected score given by user as groundtruth.
+        :return the evaluation metric value for the request document.
+        """
+        actual_at_k = actual[:self.eval_at]
+        desired_at_k = actual[:self.eval_at]
```
JoanFM, line 23: desired_at_k should be `desired[:self.eval_at]`?

JoanFM (changes requested):

evaluators/rank/NdcgEvaluator/__init__.py
JoanFM: Or move it outside the class, as a helper function somewhere

evaluators/rank/NdcgEvaluator/mainfest.yml
```diff
@@ -0,0 +1,12 @@
...
+kind: evaluator
+type: pod
+description: evaluates normalized discounted cumulative gain for information retrieval.
+author: Jina AI Dev-Team (dev-team@jina.ai)
+url: https://jina.ai
+vendor: Jina AI Limited
+documentation: https://github.com/jina-ai/jina-hub
+version: 0.0.1
+license: apache-2.0
+keywords: [ndcg, normalized discounted cumulative gain, information retrieval evaluation]
```
bhavsarpratik, line 12: Should we have word `evaluator` in the keywords?

JoanFM: yes good point!

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+        if len(actual) < 2:
+            raise ValueError(f'Expecting gains with minimal length of 2, {len(actual)} received.')
+        dcg = self._compute_dcg(gains=actual_at_k)
+        idcg = self._compute_idcg(gains=desired_at_k)
+        if idcg == 0.0:
+            return 0.0
+        else:
+            return dcg/idcg
```
bhavsarpratik, line 31:
```suggestion
        return 0.0 if idcg == 0.0 else dcg / idcg
```

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _compute_dcg(self, gains):
+        """Compute discounted cumulative gain."""
+        ret = 0.0
+        for score, position in zip(gains[1:], range(2, len(gains) + 1)):
+            ret += score/log(position, 2)
```
bhavsarpratik, line 37: Some people use `2^score -1` instead of `score`.
https://en.wikipedia.org/wiki/Discounted_cumulative_gain

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: ah!! @JoanFM good catch!

evaluators/rank/NdcgEvaluator/__init__.py
bhavsarpratik, line 37:
```suggestion
        for position, score in enumerate(gains[1:]):
            ret += score/log(position+2, 2)
```

evaluators/rank/NdcgEvaluator/__init__.py
bhavsarpratik: Unintended consequences of copy-pasting code :P

evaluators/rank/NdcgEvaluator/__init__.py
JoanFM: yes, but this implementation lets them control it on their end if the want to change that and do the power in a crafter or something

bhavsarpratik: Since NDCG is going to be the most used metric, I think we can make this easy to user by exposing an argument which controls whether to use `score` or `2^score - 1`. This is actually quite subtle and our users might not notice. By default, we can go with `2^score - 1`

JoanFM: it is a valid option.

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: i think it make more sense to leave it inside the class, logically it should be inside.

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: it's my personal hobbit to not write if else in one line, but change it also fine

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
bwanglzu: done

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: done

bwanglzu: I updated the PR with 2 functions, I'll leave the code as same because `log(position+2)` makes the code a bit hard to read.

bwanglzu:
@nan-wang @JoanFM @bhavsarpratik PR updated based on your reviews (accepted most of the comments), please take a look again.

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -8,15 +8,18 @@ def evaluator():
...
-    ([.4, .1, .0], 0.396)
-])
+@pytest.mark.parametrize('actual, expected',
+    [
+        ([.4, .1, .8], 0.795),
+        ([.0, .1, .4], 0.279),
+        ([.4, .1, .0], 0.396),
+    ]
+)
 def test_encode_success(evaluator, actual, desired, expected):
```
bhavsarpratik: I am bit confused about the usage of 3 things actual, desired and expected. Can we have only actual score and expected score?

bhavsarpratik: `test_encode` seems incorrect

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -8,15 +8,18 @@ def evaluator():
...
-    assert evaluator.evaluate(actual=actual, desired=desired) == pytest.approx(expected)
+    assert evaluator.evaluate(actual=actual, desired=desired) == pytest.approx(expected, 0.1)
 
 def test_encode_fail(evaluator, desired):
```
bhavsarpratik, line 30: `test_encode` seems incorrect

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @property
+    def metric(self):
+        return f'nDCG@{self.eval_at}'
+
+    def evaluate(
+            self,
+            actual: Sequence[Any],
+            desired: Sequence[Any],
+            use_traditional_formula=False,
```
JoanFM, line 19: I do not like this name, seems like I need to know history to use this executor.

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+        ret = 0.0
+        if use_traditional_formula:
+            for score, position in zip(gains[1:], range(2, len(gains) + 1)):
+                ret += score/log(position, 2)
+            return gains[0] + ret
+        for score, position in zip(gains, range(1, len(gains) + 1)):
+            ret += (pow(2, score) - 1)/log(position + 1, 2)
+        return ret
+
+    def _compute_idcg(self, gains, use_traditional_formula):
```
JoanFM, line 49: Put these functions outside the class shows better that they do not depend on parameters.

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -0,0 +1,32 @@
...
+def evaluator():
+    return NDCGEvaluator(eval_at=3)
+
+@pytest.fixture(scope='function')
+def desired():
+    return [.8, .4, .1, 0]
+
+@pytest.mark.parametrize('actual, use_traditional_formula, expected',
+    [
+        ([.4, .1, .8], True, 0.795),
```
JoanFM, line 15: Missing lots of edge cases, what if one of the arrays is of length 0? What if eval at is 0? What if eval at is larger than any of the arrays? Desired should also be parametrized and not a fixture.
Test with different lengths of actual and desired

JoanFM (changes requested):
Missing more careful testing.

evaluators/rank/NdcgEvaluator/README.md
```diff
@@ -0,0 +1,4 @@
+# NDCGEvaluator
+
+`NDCGEvaluator` measures the performance of a retrieval system based on the graded relevance of the predicted scores and desired scores. It ranges from `0.0` to `1.0`, with `1.0` representing the ideal ranking of the retrieved result.
```
JoanFM, line 3: Add at the end.

`Make sure to adapt `id_tag` from RankEvaluateDriver to a field in document tags containing the relevance scores`

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -0,0 +1,32 @@
...
+        ([.0, .1, .4], False, 0.210),
+        ([.4, .1, .0], False, 0.396),
+        ([.0, .0, .0], True, 0.0), # no match
+        ([.0, .0, .0], False, 0.0),
+        ([.8, .4, .1], True, 1.0), # all match
+        ([.8, .4, .1], False, 1.0),
+    ]
+)
+def test_encode(evaluator, actual, desired, use_traditional_formula, expected):
+    assert evaluator.evaluate(actual=actual, desired=desired, use_traditional_formula=use_traditional_formula) == pytest.approx(expected, 0.1)
```
JoanFM, line 28: 0.1 seems like a too large approximation

JoanFM (changes requested):

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: any suggestions?

JoanFM: Power_relevance?

bwanglzu: I agree.. `use_traditional_formula ` sounds wired.. since we don't really have a name for these two types of formulars.. some poeple use `e` as log base, some people use `2`. A lot of things were not standardised. @JoanFM

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -0,0 +1,38 @@
...
+    assert evaluator.evaluate(
+        actual=actual,
+        desired=[.8, .4, .1, 0],
+        power_relevance=power_relevance
+    ) == pytest.approx(expected, 0.01)
+
+@pytest.mark.parametrize('actual', [
+    ([]), ([0.8])
+])
+def test_encode_fail(evaluator, actual):
```
JoanFM, line 36: What happens if desired is empty?

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+
+from .. import NDCGEvaluator
+
+@pytest.fixture(scope='function')
+def evaluator():
+    return NDCGEvaluator(eval_at=3)
+
+@pytest.mark.parametrize('actual, evaluator, power_relevance, expected',
```
JoanFM, line 9: What happens if negative score values are passed?

JoanFM (changes requested):

JoanFM:
Good work! Just missing some edge cases to be tested

evaluators/rank/NdcgEvaluator/tests/test_ndcg_evaluator.py
JoanFM: expected here is the evaluation result expected, not the expected score

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @property
+    def metric(self):
+        return f'nDCG@{self.eval_at}'
+
+    def evaluate(
+            self,
+            actual: Sequence[Any],
+            desired: Sequence[Any],
+            power_relevance=True,
```
JoanFM, line 37: I did not see it, `power_relevance` should be a parameter of the class but not something in `evaluate` signature.

This class will be driven by a `driver` that will just consider the `BaseEvaluator` interface and will never provide an extra parameter for this.

JoanFM (changes requested):

evaluators/rank/NdcgEvaluator/__init__.py
JoanFM: the signature is not real, if I pass `str` it won't be able to power them, what about `Sequence[float]` or `Sequence[Union[int, float]]`?

JoanFM (changes requested):

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -45,10 +45,18 @@ def evaluate(
             For detailed information, please check https://en.wikipedia.org/wiki/Discounted_cumulative_gain
         :return the evaluation metric value for the request document.
         """
+        # Information gain must be greater or equal to 0.
+        if any(item < 0 for item in actual) or any(item < 0 for item in desired):
+            raise ValueError('One or multiple score is less than 0.')
+        # Desired must in desc order.
+        if desired != sorted(desired, reverse=True):
```
JoanFM, line 52: I think this is too expensive, I think we just add in docstrinh as note that it assumes that desired come ordered?

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -45,10 +45,18 @@ def evaluate(
...
+            raise ValueError('Information gain of desired must in a desc order.')
         actual_at_k = actual[:self.eval_at]
         desired_at_k = desired[:self.eval_at]
-        if len(actual) < 2:
-            raise ValueError(f'Expecting gains with minimal length of 2, {len(actual)} received.')
+        if len(actual_at_k) < 2:
```
JoanFM, line 56: Why it has to be larger than 2? Is the formula not defined for that case?

JoanFM (commented):
Remember to add the power relevance atribute in constructor.

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: ok

evaluators/rank/NdcgEvaluator/__init__.py
bwanglzu: yes,  it's useless to check ndcg with one value, right? the score always equal to the gain itself.

JoanFM: It is useless but not invalid, why make it an exception? To me exception is something really exceptional no?

evaluators/rank/NdcgEvaluator/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+def _compute_idcg(gains, power_relevance):
+    """Compute ideal discounted cumulative gain."""
+    sorted_gains = sorted(gains, reverse=True)
+    return _compute_dcg(sorted_gains, power_relevance)
+
+class NDCGEvaluator(BaseRankingEvaluator):
+    """
+    :class:`NDCGEvaluator` evaluates normalized discounted cumulative gain for information retrieval.
+    """
+    def __init__(self, eval_at, power_relevance=True):
```
JoanFM, line 28: In order for the tests to pass we need to pass a default value to eval at.

JoanFM (approved):

PR #462, opened 2020-10-29
feat: add bleu evaluator

theUnkownName:


evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,38 @@
+from jina.executors.evaluators import BaseEvaluator
+from typing import Sequence, Any
+import nltk.translate.bleu_score as bleu
```
JoanFM, line 3: import these dependencies in the evaluate method

JoanFM: Try following standard when importing.

1. System lib dependencies
Space
2. External imports
Space
3. Imports from jina

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    
+    
+    def get_nltk_bleu_score(self, hypothesis, reference):
+        return bleu.sentence_bleu(reference, hypothesis)
+
+
+    def evaluate(self,
+            hypothesis,
+            reference,
+            nltk_default = True,
```
JoanFM, line 26: nltk_default is not used. and if it was used it should be a parameter and introduced in the __init__. Learning, be very careful with changing the interface of actual executors methods.

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+            **kwargs) -> float:
+        """"
+        :param actual: the scores predicted by the search system.
+        :param desired: the expected score given by user as groundtruth.
+        :return the evaluation metric value for the request document.
+        """
+
+        BLEUScore = self.get_nltk_bleu_score(hypothesis, reference)
+
+        return BLEUScore
```
JoanFM, line 37: directly `return self.get_nltk_bleu_score(hypothesis, reference)`

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    A perfect match will score 1.0 and a perfect unmatch will score 0.0
+    """
+    
+    
+    def get_nltk_bleu_score(self, hypothesis, reference):
+        return bleu.sentence_bleu(reference, hypothesis)
+
+
+    def evaluate(self,
+            hypothesis,
```
JoanFM, line 24: call it `actual` , `desired`

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -0,0 +1,16 @@
+from .. import BleuEvaluator
+import pytest
+
+
+def test_bleu_evaluator():
+    reference = [ ['All', 'cats', 'are', 'beautiful'], ['All', 'cats', 'are', 'cute'] ]
```
JoanFM, line 6: It is weird, I think what the encoder should see is only `strings` and not `arrays of strings`. This is gonna be used to evaluate texts from crafters. Another thing is wether inside we may need to cut it.

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: Oh yes, I was trying something different and forgot to remove this. Ill take it out

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
theUnkownName: To be clear, this will compare only one string to another one? in this example, the reference (desired) will be a list of strings or will also be only one string?

JoanFM: As per my understanding we should have :

`BleuEvaluator.evaluate('All cats are beautiful', 'All cats are cute')`

JoanFM: This will compare one-by-one a document `jina_pb2.Document.text` vs a  groudntruth `jina_pb2.Document.text`

JoanFM: Also try to add tests with edge cases, this can serve as documentation on how things work. Like what happens with `lower` and `upper` cases, what happens if the `desired` is empty or if the `actual` is empty ...

theUnkownName: Yeah sure, it has still many tests missing, this was just the first draft to check everything was in order

JoanFM: yes, just reminder ...

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -1,38 +1,38 @@
...
     """
     
     
-    def get_nltk_bleu_score(self, hypothesis, reference):
-        return bleu.sentence_bleu(reference, hypothesis)
+    def get_nltk_bleu_score(self, desired, actual):
+        return bleu.sentence_bleu(desired, actual)
 
 
     def evaluate(self,
```
JoanFM, line 18: change the order, it is `actual` and then `desired`

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -1,38 +1,38 @@
...
-            hypothesis,
-            reference,
-            nltk_default = True,
+            desired,
+            actual, 
             *args,
             **kwargs) -> float:
         """"
-        :param actual: the scores predicted by the search system.
         :param desired: the expected score given by user as groundtruth.
```
JoanFM, line 24: here you are not talking about scores but `text`

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -1,38 +1,38 @@
...
+        NLTK expectes an array of strings, 
+        so the incoming string needs to be tokenized first.
+        They will be stored in a desired_list
+        """
+        
+        # tokenize
+        desired_list = desired.split()
+        actual_list = actual.split()
+        
+        return self.get_nltk_bleu_score([desired_list], actual_list)
```
JoanFM, line 37: do you need to make a `list` of a `list`?

JoanFM (changes requested):

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: Where you mean is it in that order? in the nltk lib?

JoanFM: The interface of the BaseEvaluator, and this is what the driver will provide.

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: Yep, cause it can be a list of many references, and each of them is a list of strings (one str per word)

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: Ahhh right, missed it

evaluators/text/BleuEvaluator/__init__.py
JoanFM: Ok!

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -22,4 +22,9 @@ def test_bleu_evaluator_caps():
     actual = 'ALL CATS ARE SO SUPER BEAUTIFUL'
     assert BleuEvaluator().evaluate(actual, desired) == 1.0
 
+def test_bleu_evaluator_fail():
+    desired = 'All cats are so super beautiful'
+    actual = 'Some dogs are also cute memes'
+    with pytest.raises(Exception):
```
JoanFM, line 28: why does it raise an exception?

JoanFM (changes requested):

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
theUnkownName: Bleu is checking how similar `desired` and `actual`. Since those two are not really similar, it should fail, shouldn't it?

JoanFM: no no, it should return a value smaller than one, but not raise an Exception.

theUnkownName: The test was checking if it was 1, otherwise, it would raise the exception, but sure, let me change it

JoanFM: ah ok!!! I understand, this is not a good testing style, u should test `assert result < 1.0` or even better compute it offline and add the test there

theUnkownName: Oh ok, got it

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,37 @@
+from jina.executors.evaluators import BaseEvaluator
+
+class BleuEvaluator(BaseEvaluator):
```
JoanFM, line 3: Should inherit BaseTextEvaluator

JoanFM (changes requested):

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -0,0 +1,30 @@
...
+    desired = 'All cats are so super beautiful'
+    actual = ''
+    assert BleuEvaluator().evaluate(actual, desired) == 0.0
+
+def test_bleu_evaluator_desired_emtpy():
+    desired = ''
+    actual = 'All cats are so super beautiful'
+    assert BleuEvaluator().evaluate(actual, desired) == 0.0
+
+def test_bleu_evaluator_caps():
```
JoanFM, line 20: Try to use pytest parametrization for these tests. Look at the other PRs on evaluators as inspiration

theUnkownName: Got it!

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: Yeah right, changed it

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -25,6 +25,5 @@ def test_bleu_evaluator_caps():
 def test_bleu_evaluator_fail():
     desired = 'All cats are so super beautiful'
     actual = 'Some dogs are also cute memes'
-    with pytest.raises(Exception):
-        assert BleuEvaluator().evaluate(actual, desired) == 1.0
+    assert BleuEvaluator().evaluate(actual, desired) == 0.7351508268457205
```
JoanFM: use some `almost equal` or some approximation, this test is very likely to fail in some platforms. we do not need this much precision for a test.

JoanFM: check `pytest.approx`

theUnkownName: Got it

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    Evaluates the generated sentence (actual) against a desired sentence. 
+    It will use the Bleu on NLTK package.
+    A perfect match will score 1.0 and a perfect unmatch will score 0.0
+    """
+    
+    def get_nltk_bleu_score(self, desired, actual):
+        import nltk.translate.bleu_score as bleu    
+        from nltk.translate.bleu_score import SmoothingFunction
+
+        return bleu.sentence_bleu(desired, actual, smoothing_function=SmoothingFunction().method4)
```
JoanFM, line 15: according to the `README` this `method4` should only be used if `n-grams` of less than `4`? Shouldn't this be an option as input?

theUnkownName: To have an input of the` n-gram` size?

JoanFM: to use or not the `smoothing_function`. It just feels incoherent the documentation message and the implementation, why are we forcing the usage of this smoothing function without making sure that the checks to activate it are valid?

I would just give the option to the user to use or not the smoothing function.

JoanFM: u can ask for an input argument (`smoothing_method`) and keep a static map of `str`  -> `SmoothingFunction().method` or even use the `getattr(SmoothingFunction(), method)` (which is this reflexion we want to avoid, but maybe in this case may be justified)

theUnkownName: Yeah makes sense, I also just saw another thing that should improve the scoring, ill change some stuff

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -7,17 +7,48 @@ class BleuEvaluator(BaseTextEvaluator):
...
+        if overlaps == 4:
+            return bleu.sentence_bleu(desired, actual, weights = (0, 0, 0, 1), smoothing_function=SmoothingFunction().method4)
+        else:
+            return bleu.sentence_bleu(desired, actual) #otherwise use standard weights (0.25, 0.25, 0.25, 0.25) and no Smoothin function
 
 
     def evaluate(self,
             actual,
             desired,
+            max_order=4,
```
JoanFM, line 50: These parameters do not belong here, they may be part of  the `__init__`. but please do not change the `interface`

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -8,7 +8,7 @@
         ('All cats are so super beautiful', '', 0.0),
         ('', 'All cats are so super beautiful', 0.0),
```
JoanFM, line 9: With the new additions we need to check that it works with any `ngrams` count

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -7,17 +7,48 @@ class BleuEvaluator(BaseTextEvaluator):
...
+
     
-    def get_nltk_bleu_score(self, desired, actual):
+    def get_nltk_bleu_score(self, desired, actual, overlap):
         import nltk.translate.bleu_score as bleu    
         from nltk.translate.bleu_score import SmoothingFunction
 
-        return bleu.sentence_bleu(desired, actual, smoothing_function=SmoothingFunction().method4)
+        #Check the n-gram and reset the weights accordingly
+        overlaps = sum(overlap.values())
```
JoanFM, line 34: is there any reference that explains this logic here?

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -7,17 +7,48 @@ class BleuEvaluator(BaseTextEvaluator):
     It will use the Bleu on NLTK package.
     A perfect match will score 1.0 and a perfect unmatch will score 0.0
     """
+
+    def count_ngram(self, text, max_ngram):
```
JoanFM, line 11: this can be a `staticmethod` as it does not require `self` for anything

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -7,17 +7,48 @@ class BleuEvaluator(BaseTextEvaluator):
...
+        for n_order in range(1, max_ngram):
+            for i in range(len(text) - n_order):
+                ngram = tuple(text[i:i+n_order])
+                counter[ngram] += 1
+        
+        return counter
+
     
-    def get_nltk_bleu_score(self, desired, actual):
+    def get_nltk_bleu_score(self, desired, actual, overlap):
```
JoanFM, line 29: same here, it can be a `static` method

JoanFM (changes requested):

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: What you think of how it's done now? I think it's maybe better to check the n-gram and then decide how to process it instead of getting input from the user?

JoanFM: if there is a strong reference for this, I can agree

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: I added an article about it

evaluators/text/BleuEvaluator/tests/test_bleuevaluator.py
```diff
@@ -0,0 +1,25 @@
...
+        ('A small cat', 'A small cat', 1.0),
+        ('Even smaller', 'Even smaller', 1.0)
+    ])
+
+
+def test_bleu_evaluator(actual, desired, score):
+    evaluator = BleuEvaluator()
+    assert evaluator.evaluate(actual, desired) == pytest.approx(score, 0.1)
+
+
```
JoanFM, line 22: add individual tests for `count_ngram` so that is clear what it does

JoanFM (changes requested):

evaluators/text/BleuEvaluator/README.md
```diff
@@ -0,0 +1,6 @@
+# BleuEvaluator
+
+Bilingual Evaluation Understudy Score. Evaluates the generated sentence against a desired sentence. 
+A perfect match will score 1.0 and a perfect unmatch will score 0.0
+
+BLEU works well with n-grams of at least 4, if less than that, it's neccessary to use a smoothing functionå.
```
cristianmtr, line 6:
```suggestion
BLEU works well with n-grams of at least 4, if less than that, it's neccessary to use a smoothing function.
```

Remove trailing `å`

evaluators/text/BleuEvaluator/README.md
```diff
@@ -0,0 +1,6 @@
+# BleuEvaluator
+
+Bilingual Evaluation Understudy Score. Evaluates the generated sentence against a desired sentence. 
+A perfect match will score 1.0 and a perfect unmatch will score 0.0
```
cristianmtr, line 4: "perfect unmatch" sounds odd. "perfect mismatch", maybe?

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,83 @@
+from jina.executors.evaluators.text import BaseTextEvaluator
+
+class BleuEvaluator(BaseTextEvaluator):
+    """
+    :class:`BleuEvaluator`Bilingual Evaluation Understudy Score. 
+    Evaluates the generated sentence (actual) against a desired sentence. 
+    It will use the Bleu on NLTK package.
+    A perfect match will score 1.0 and a perfect unmatch will score 0.0
```
cristianmtr, line 8: Same as above

evaluators/text/BleuEvaluator/README.md
theUnkownName: So many typos

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -70,14 +57,5 @@ def evaluate(self,
         # set everything to undercase and tokenize
         desired_list = desired.lower().split()
         actual_list = actual.lower().split()
+        return self.get_score([desired_list], actual_list, len([desired_list]))
```
JoanFM, line 60: try this:

```
>>> desired = "hey I am here"
>>> desired_list = desired.lower().split()
>>> desired_list
['hey', 'i', 'am', 'here']
>>> len([desired_list])
1
>>> 
```
len([desired_list]) will always return 1. Why are the `n_grams` computed from the `desired_list` and not the `actual_list`?

JoanFM (changes requested):

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: You're right, it was checking the list of the list. 
I changed the second test to use a n-gram less than 4, so we see the difference between resetting the weights or not.

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+    @staticmethod
+    def get_score(actual_list, desired_list, n_gram):
+        '''
+        Cumulative score is the calculation of individual n-grams
+        from 1 to n-order, and then weights them with the geometric mean
+        - desired_list, actual_list: are the sentences to be scored
```
JoanFM, line 25: have a standard `docstring` with the proper `params`. I do not understand `It will check the biggest n_gram possible` if the `n_gram` is an input to the function.

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+        NLTK expectes an array of strings, 
+        so the incoming string needs to be tokenized first.
+        They will be stored in a desired_list and actual_list accordingly
+        """
+
+        # set everything to undercase and tokenize
+        desired_list = desired.lower().split()
+        actual_list = actual.lower().split()
+
+        return self.get_score(actual_list, [desired_list], len(actual_list))
```
JoanFM, line 61: why now it changed, and the n_grams are considered from the `actual_list`?

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    https://machinelearningmastery.com/calculate-bleu-score-for-text-python/
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+    @staticmethod
+    def get_score(actual_list, desired_list, n_gram):
```
JoanFM, line 21: add type hinting, so that one can see what types are entering, they seem to be of the same type, but for the calling point of view seems that one is a list and the other a List of Lists

evaluators/text/BleuEvaluator/__init__.py
JoanFM: According to `bleu` these are the parameters: 

```
  :param references: reference sentences
    :type references: list(list(str))
    :param hypothesis: a hypothesis sentence
    :type hypothesis: list(str)
```

JoanFM (changes requested):

evaluators/text/BleuEvaluator/__init__.py
theUnkownName: The `n_gram` needs to be from the `actual_list` since that is the `candidate` with NLKT, if our `actual` were 2 words only, it wouldn't matter if the `desired` is bigger, since we would only have a a 2-gram.
It changed because it was an error before, I was passing the list of all the possible lists of `desired`, since we only have one element in the list, it didn't matter as much, but I keep it as standard with most use-cases with NLTK

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -1,5 +1,8 @@
 from jina.executors.evaluators.text import BaseTextEvaluator
 
+from typing import List
```
JoanFM, line 3: keep the standard import order

evaluators/text/BleuEvaluator/__init__.py
```diff
@@ -18,12 +21,14 @@ def __init__(self, *args, **kwargs):
 
 
     @staticmethod
-    def get_score(actual_list, desired_list, n_gram):
+    def get_score(actual_list: List[str], desired_list: List[List[str]], n_gram):
         '''
+        - desired_list: A List of a List of all possible sentences ex: [['cats are cute'], ['dogs are cute']]
```
JoanFM, line 26: look how the docstring is created, with the ```:params ```. Also add hint for `n_gram`

evaluators/text/BleuEvaluator/README.md
```diff
@@ -0,0 +1,18 @@
+# BleuEvaluator
+
+Bilingual Evaluation Understudy Score. Evaluates the generated sentence against a desired sentence. 
+A perfect match will score 1.0 and a complete mismatch will score 0.0
+
+BLEU works well with n-grams of at least 4, if less than that, it's neccessary to use a smoothing function and reset the weights.
+If this is not done, it can be the case that we get results smaller than zero which shouldn't happen, as well as returning the error
+
+```
+ZeroDivisionError: Fraction(0, 0)
```
JoanFM, line 10: Do we have a case where we raise this exception? if so , we should test it

theUnkownName: It won't get there because we are setting the weights according to the size of the n_gram, so it will be checked in `get_score` and use only the correct weights

JoanFM: So better remove the comment?

theUnkownName: It was more to explain why I was doing the checks, but sure you're right, it makes no sense

JoanFM (approved):

PR #464, opened 2020-10-31
feat: add edit distance evaluator

bhavsarpratik:
Add edit distance evaluator


evaluators/text/EditDistanceEvaluator/tests/test_editdistanceevaluator.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from .. import EditDistanceEvaluator
+
+
+@pytest.mark.parametrize('actual, desired, distance',
```
JoanFM, line 6: What happens when mixed strings of different lengths not empty and also capital letters?

JoanFM (changes requested):
Just some edge cases

evaluators/text/EditDistanceEvaluator/tests/test_editdistanceevaluator.py
JoanFM: What if sentences with punctuation marks?

JoanFM (approved):

PR #466, opened 2020-11-01
feat: add reciprocal rank evaluator

bhavsarpratik:


evaluators/text/ReciprocalRankEvaluator/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def metric(self):
+        return f'ReciprocalRank@{self.eval_at}'
+
+    def evaluate(self, actual: Sequence[str], desired: Sequence[str], *args, **kwargs) -> float:
```
JoanFM, line 17: why are they `Sequence[str]`?

evaluators/text/ReciprocalRankEvaluator/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+        if len(actual)==0 or len(desired)==0:
+            return 0.0
+        try:
+            return float(1/(actual[:self.eval_at].index(desired[0])+1))
```
JoanFM, line 21: in which case there would be an exception? in case of a divided by 0?

evaluators/text/ReciprocalRankEvaluator/tests/test_reciprocalrankevaluator.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+from .. import ReciprocalRankEvaluator
+
+
+@pytest.mark.parametrize('actual, desired, score',
+[
+    ([],[], 0.0),
+    (['hey', 'hello', 'hi', 'heya'],[], 0.0),
```
JoanFM, line 9: I think these tests are a little missleading. Normally uses `RankingEvaluations` to evaluate a list of `documentIds` or `scores`. I guess `hey, hello, hi...` can be `document identifiers` but if I come just to see the `tests` I get very confused.

JoanFM (changes requested):

evaluators/text/ReciprocalRankEvaluator/__init__.py
bhavsarpratik: Only in the case of desired[0] is not found in `actual[:self.eval_at]`

evaluators/text/ReciprocalRankEvaluator/tests/test_reciprocalrankevaluator.py
bhavsarpratik: Okay, I will make changes for `documentIds`. I was considering doc content as the `actual` and `desired`

evaluators/text/ReciprocalRankEvaluator/__init__.py
```diff
@@ -14,10 +14,18 @@ def __init__(self, *args, **kwargs):
     def metric(self):
         return f'ReciprocalRank@{self.eval_at}'
 
-    def evaluate(self, actual: Sequence[str], desired: Sequence[str], *args, **kwargs) -> float:
+    def evaluate(self, actual: Sequence[int], desired: Sequence[int], *args, **kwargs) -> float:
```
JoanFM, line 17: Signature change to Sequence[Any] I think is more appropiate

JoanFM (changes requested):

evaluators/text/ReciprocalRankEvaluator/__init__.py
bhavsarpratik: This executor supports only document IDs which will be int. It doesn't support scores which will be float. Do we still want to change it to Any?

JoanFM: DocumentIDs may come in form of str, u can do Sequence[Union[str, int]]

evaluators/rank/ReciprocalRankEvaluator/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+        super().__init__(*args, **kwargs)
+
+    @property
+    def metric(self):
+        return f'ReciprocalRank@{self.eval_at}'
+
+    def evaluate(self, actual: Sequence[Union[str,int]], desired: Sequence[Union[str,int]], *args, **kwargs) -> float:
+        """
+        Args:
+            actual (Sequence[int]): should be a sequence of document IDs
```
JoanFM, line 20: look at the syntax used to describe the params in other classes.

evaluators/rank/ReciprocalRankEvaluator/__init__.py
```diff
@@ -0,0 +1,31 @@
+from typing import Sequence, Union
+
+from jina.executors.evaluators.rank import BaseRankingEvaluator
+
+class ReciprocalRankEvaluator(BaseRankingEvaluator):
+    """
+    :class:`ReciprocalRankEvaluator` Gives score as per reciprocal rank metric.
+    """
+
+    def __init__(self, *args, **kwargs):
```
JoanFM, line 10: We need to see if we want to put the `eval_at` default here or in the `config.yml`. I would put it here and pass it to the `super().__init__()` call

JoanFM (approved):

PR #469, opened 2020-11-02
feat: add f1 score eval

pgiank28:


JoanFM:
recheckcla

evaluators/rank/f1ScoreEvaluator/__init__.py
```diff
@@ -0,0 +1,39 @@
+from jina.executors.evaluators.rank import BaseRankingEvaluator
+
+class f1ScoreEvaluator(BaseRankingEvaluator):
+    """
+    :class:`f1ScoreEvaluator` Gives the f1 score of a result given the groundtruth..
+    """
+
+    def __init__(self, eval_at ,*args, **kwargs):
```
JoanFM, line 8: we need a default value to pass the tests

JoanFM: it may need to be moved to the base class but put it here for now

JoanFM (changes requested):

JoanFM:
no need to close it, u just need to sign the `CLA` agreement

pgiank28:

evaluators/rank/f1ScoreEvaluator/__init__.py
JoanFM: I think 1 would make a better default

pgiank28: probably,but can you explain ?

JoanFM: Assuming that it should not be defined here, we may address this in another PR. We need to provide a k to do an eval at. 0 seems like the last eval at point someone would use because it does not involve any evaluation actually. Don't you think so?

pgiank28: you're right

JoanFM (approved):

PR #471, opened 2020-11-04
feat: add infinite norm as embedding evaluator

rutujasurve94:


evaluators/embedding/InfiniteNormEvaluator/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+        actual = expand_vector(actual)
+        desired = expand_vector(desired)
+        return _infinitenorm(actual, desired)
+
+
+def _get_ones(x, y):
+    return np.ones((x, y))
+
+
+def _ext_A(A):
```
JoanFM, line 30: remove all the unused methods.

JoanFM (changes requested):

evaluators/embedding/InfiniteNormEvaluator/__init__.py
rutujasurve94: Keeping this as it will required to calculated norms on transformed `np.array` A

JoanFM (approved):

PR #474, opened 2020-11-04
feat(evaluator): add average precision

kilsenp:
New PR. Closes jina-ai/jina#1216.

Adding Average Precision (AP) as an evaluator.
See tests for some sanity checks.

I considered adding this to jina-hub, but thought this can fit here as well as there is no other dependencies besides numpy.

The BaseRankingEvaluator class expects a eval_at argument which is not necessary for AP as it evaluates on the whole ranking.
It could be used to limit the AP to only consider the first x results though.


evaluators/rank/AveragePrecision/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+        :return the evaluation metric value for the request document
+        """
+        matches = []
+        for idx, doc_id in enumerate(desired):
+            matches.append(doc_id in actual[:idx+1])
+        matches = np.array(matches)
+        cumsum = np.cumsum(matches)
+        precision = cumsum / np.arange(1, len(matches) + 1)
+
+        # Does not match behaviour in
```
JoanFM, line 28: Does this mean we have a bug in our PrecisionEvaluatorm

evaluators/rank/AveragePrecision/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jina>=0.7
+numpy
```
JoanFM, line 2: Numpy there is no need to add since is one of our core dependencies.

evaluators/rank/AveragePrecision/__init__.py
```diff
@@ -0,0 +1,38 @@
+from typing import Sequence, Any
+import numpy as np
+from jina.executors.evaluators.rank import BaseRankingEvaluator
+
+
+class AveragePrecisionEvaluator(BaseRankingEvaluator):
```
JoanFM, line 6: I know it may be design fault ( we will change it) but for now we will need to have a default value of eval at in an __init__ in order to pass our CI.

JoanFM (commented):
Beyond the comment on the init, the CI won't pass due to an issue we found that we are working on

evaluators/rank/AveragePrecision/__init__.py
kilsenp, line 28: Not sure if you want to call it a bug as it's up for definition.
This precision array basically contains all precision@x where x is the idx of the array.
In the array here precision@0 is set to 1.0, whereas in the precision evaluator precision@0 will return 0.
Precision here must be set to 1.0 as the integral will otherwise return other results.
Most notably the test case desired=[0, 1, 2, 3] actual=[0, 1, 2, 3] would no longer return 1.0.

JoanFM: Ah okey nice explanation

kilsenp: Not sure if you want to call it a bug as it's up for definition.
This precision array basically contains all precision@x where x is the idx of the array.
In the array here precision@0 is set to 1.0, whereas in the precision evaluator precision@0 will return 0.
Precision here must be set to 1.0 as the integral will otherwise return other results.
Most notably the test case desired=[0, 1, 2, 3] actual=[0, 1, 2, 3] would no longer return 1.0.

evaluators/rank/AveragePrecision/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    """A :class:`AveragePrecisionEvaluator` evaluates the Average Precision of the search.
+       https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Average_precision
+    """
+    def __init__(self, eval_at=-1, *args, **kwargs):
+        super().__init__(eval_at, *args, **kwargs)
```
nan-wang, line 11:
```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(-1, *args, **kwargs)
```

JoanFM:
Hey @kilsenp ,

there is an error that the name of the `Executor` in your `config.yml` does not match the real class name.

JoanFM (changes requested):
Adapt config and Readme to real class name

JoanFM (approved):

JoanFM:
Unittests seem not to be passing

JoanFM (approved):

PR #479, opened 2020-11-06
feat(image_reader): allow to read numpy images

florian-hoenicke:
- ndarray can be used as input for ImageReader
- ndarray without color dimensions can be used
- if there is an alpha channel, it will be converted to rgb
- raises exception in case the input array type is not uint8
- raises exception in case it has an incorrect shape
- unit tests added

### DIFF
--- a/crafters/image/ImageReader/__init__.py
+++ b/crafters/image/ImageReader/__init__.py
@@ -11,12 +11,12 @@ class ImageReader(BaseCrafter):
     shape (h,w,depth).
     """
 
-    def __init__(self, color_channel_axis: int = -1, *args, **kwargs):
+    def __init__(self, channel_axis: int = -1, *args, **kwargs):
         """
-        :param color_channel_axis: the axis id of the color channel -1 indicates the color channel info at the last axis
+        :param channel_axis: the axis id of the color channel -1 indicates the color channel info at the last axis
         """
         super().__init__(*args, **kwargs)
-        self.channel_axis = color_channel_axis
+        self.channel_axis = channel_axis
 
     def craft(self, buffer: bytes, uri: str, blob: 'np.ndarray', *args, **kwargs) -> Dict:
         """
--- a/crafters/image/ImageReader/__init__.py
+++ b/crafters/image/ImageReader/__init__.py
@@ -7,36 +7,50 @@ from jina.executors.crafters import BaseCrafter
 
 class ImageReader(BaseCrafter):
     """
-    :class:`ImageReader` loads the image from the given file path and save the `ndarray` of the image in the Document.
+    :class:`ImageReader` loads the image from the given file path, bytes or ndarray, and emits the unified `ndarray` with
+    shape (h,w,depth).
     """
 
-    def __init__(self, channel_axis: int = -1, *args, **kwargs):
+    def __init__(self, color_channel_axis: int = -1, *args, **kwargs):
         """
-        :class:`ImageReader` load an image file and craft into image matrix.
-
-        :param channel_axis: the axis id of the color channel, ``-1`` indicates the color channel info at the last axis
+        :param color_channel_axis: the axis id of the color channel -1 indicates the color channel info at the last axis
         """
         super().__init__(*args, **kwargs)
-        self.channel_axis = channel_axis
+        self.channel_axis = color_channel_axis
 
-    def craft(self, buffer: bytes, uri: str, *args, **kwargs) -> Dict:
+    def craft(self, buffer: bytes, uri: str, blob: 'np.ndarray', *args, **kwargs) -> Dict:
         """
-        Read the image from the given file path that specified in `buffer` and save the `ndarray` of the image in
-            the `blob` of the document.
+        Read the image from the given `buffer`, `uri` or `blob` and saves the unified `ndarray` of the image in
+        the `blob` of the document.
 
-        :param buffer: the image in raw bytes
+        :param buffer: the encoded image in raw bytes (png, jpg)
         :param uri: the image file path
-
+        :param blob: the image as ndarray
         """
+
+        pil_image = self.get_pil_image(buffer, uri, blob)
+        pil_image = pil_image.convert('RGB')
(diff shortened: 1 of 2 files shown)


florian-hoenicke:
fyi @hanxiao

crafters/image/ImageReader/__init__.py
```diff
@@ -7,36 +7,50 @@
 
 class ImageReader(BaseCrafter):
     """
-    :class:`ImageReader` loads the image from the given file path and save the `ndarray` of the image in the Document.
+    :class:`ImageReader` loads the image from the given file path, bytes or ndarray, and emits the unified `ndarray` with
+    shape (h,w,depth).
     """
 
-    def __init__(self, channel_axis: int = -1, *args, **kwargs):
+    def __init__(self, color_channel_axis: int = -1, *args, **kwargs):
```
JoanFM, line 14: Is it possible to keep the argument name as it was? A lot of examples depend on that and it would make our life easier

JoanFM (changes requested):
I do not fully understand why would one call a Reader from a numpy array, since the point of this crafter is to obtain the numpy array representation of the image?

It seems to me that if you already have a numpy array at this stage, what you need is something else but not a Reader?

crafters/image/ImageReader/tests/test_imagereader.py
```diff
@@ -3,35 +3,95 @@
...
 
 
 def test_io_uri():
     crafter = ImageReader()
-    tmp_fn = os.path.join(crafter.current_workspace, 'test.jpeg')
-    img_size = 50
-    create_test_image(tmp_fn, size_width=img_size, size_height=img_size)
-    test_doc = crafter.craft(buffer=None, uri=tmp_fn)
-    assert test_doc['blob'].shape == (img_size, img_size, 3)
+    filename = os.path.join(crafter.current_workspace, 'test.jpeg')
```
JoanFM, line 25: It would be cool if you could use the tmpdir fixture, we are trying to slowly update our tests with better practices.

florian-hoenicke:
@JoanFM  Thanks for your fast reply! I understand what you mean. A reader should read a file from a disk/url...
However, it can decode image bytes directly as well. To me, the ImageLoader as it is, is more like a image unifier/standardizer which converts all kinds of formats into RGB ndarrays.
If you have images as numpy array, it does not necessary mean, they are in the right format (RGB).
Maybe they are of type float between 0 and 1 or they don't have a color dimension or they have an alpha channel etc...
Would you suggest that I move this to a new crafter called something like RGBConverter?
I'm curious of what you think.

crafters/image/ImageReader/tests/test_imagereader.py
```diff
@@ -3,35 +3,95 @@
...
+    assert test_doc['blob'].shape == (img_h, img_w, 3)
+
+
+def test_blob_no_color_channel():
+    crafter = ImageReader()
+    img_h = 20
+    img_w = 30
+    numpy_image = np.random.randint(0, 255, size=(img_h, img_w), dtype=np.uint8)
+    test_doc = crafter.craft(buffer=None, uri=None, blob=numpy_image)
+    assert test_doc['blob'].shape == (img_h, img_w, 3)
```
JoanFM, line 63: why does the `result` have 3 color channels then?

JoanFM:
> @JoanFM Thanks for your fast reply! I understand what you mean. A reader should read a file from a disk/url...
> However, it can decode image bytes directly as well. To me, the ImageLoader as it is, is more like a image unifier/standardizer which converts all kinds of formats into RGB ndarrays.
> If you have images as numpy array, it does not necessary mean, they are in the right format (RGB).
> Maybe they are of type float between 0 and 1 or they don't have a color dimension or they have an alpha channel etc...
> Would you suggest that I move this to a new crafter called something like RGBConverter?
> I'm curious of what you think.

I would personally move it as a numeric `crafter` that converts or manipulates arrays as images. Let's see what others think though. 

Actually I think the ImageReader would need to have the option to load grayscale images, but that is another point.

crafters/image/ImageReader/tests/test_imagereader.py
florian-hoenicke: The result is intended to be a standard format for images. Our image crafters assume a color channel dimension and the encoders like BigTransfer require ndarrays in RGB as well.
For instance, if you use the mnist dataset from keras,
`(x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()`
the image data does not contain a color channel. 
In the current implementation, an application developer has to introduce the color channel first.
If we have encoders in the future which uses just one channel or even an alpha channel, than we could provide an output mode in the the loders constructor (RGB, RGBA, GRAYSCALE...)

florian-hoenicke:
True, it would be nice to have make other image outputs formats configurable as well.
I thought about numeric crafters as well. Howerver, since it is related to images, I thought there would be a good place. Yes let's see if someone of the others enlightens us :)

crafters/image/ImageReader/tests/test_imagereader.py
JoanFM: To me this is an assumption tht someone would use these encoders, while every executor should be agnostic and independent to what others do. It is strange to me that a Reader class actually transforms so deeply my input.

JoanFM:
I would definitely suggest to create a numeric crafter for this purpose.

crafters/image/ImageReader/__init__.py
florian-hoenicke: sure, done

crafters/image/ImageReader/tests/test_imagereader.py
florian-hoenicke: nice idea, I changed it to use the tmpdir.

deepankarm:
Agreed with @JoanFM. We should have a different crafter for this.

JoanFM:
Yes, let's have this logic inside another `executor` and I think (although counterintuitive) from an `interface` point of view it should be a `numerical` crafter.

PR #481, opened 2020-11-08
feat(hub): ranker for aggregation functions

florian-hoenicke:
In our case a document is a Talent and chunks are images of the talent.
In order to have a match between two talents, it is not enough that one image is matching quite well (MinRanker).
Instead, all images have to match more or less. 
I could have created an AverageRanker or MedianRanker. But since we have a MinRanker and MaxRanker already, I bundled them into an AggregationRanker, where you can define the aggregate function in the constructor (min, max, mean, median, sum, prod).
Also, it can be defined if the resulting score has to be inverted (x=1/(1+x)) - like it is done in the MinRanker.
Would you suggest to deprecate the MinRanker and MaxRanker?
Do you even think that having all aggregate functions in one ranker is a good idea?
Creating a ranker for each of these functions seemed to me a bit overkill.


JoanFM:
I think is a very nice ranker and a good question wwther we should merge them all. I think that if we go for merging them all we should them almost deprecate Chunk2DocRanker and have this one instead. 

But I personally like better the option to split them so that they grow independently. Otherwise if a new one appears the more natural way to go would be changing this ranker, while I think it is a good thing if we can have executors being very stable

JoanFM:
Actually I was thinking for a while that Chunk2DocRanker is not a name I liked much and AggregateRanker seems like a very good name for what it does, what do u think @nan-wang @maximilianwerk ?

rankers/AggregateRanker/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+        """
+        super().__init__(*args, **kwargs)
+        self.is_reversed_score = is_reversed_score
+        self.aggregate_function = aggregate_function
+
+    def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
+        scores = match_idx[self.COL_SCORE]
+        if self.aggregate_function in self.AGGREGATE_FUNCTIONS:
+            aggregated_score = getattr(np, self.aggregate_function)(scores)
```
cristianmtr, line 27: Shouldn't this be checked at `__init__` time?

rankers/AggregateRanker/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        else:
+            raise ValueError(f'The aggregate function "{self.aggregate_function}" is not defined.')
```
cristianmtr, line 29: I would make the exception more explicit, like

`... "{self.aggregate_function}" is not one of "{self.AGGREGATE_FUNCTIONS}".')`

rankers/AggregateRanker/__init__.py
florian-hoenicke: Yes, nice catch! I fully agree. We don't want to have the check and the getattr within the loop.

rankers/AggregateRanker/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+class AggregateRanker(Chunk2DocRanker):
+    """
+    :class:`AggregateRanker` calculates the score of the matched doc form the matched chunks. For each matched doc, the
+    score is aggregated from all the matched chunks belonging to that doc.
+    """
+
+    AGGREGATE_FUNCTIONS = ['min', 'max', 'mean', 'median', 'sum', 'prod']
+
+    def __init__(self, aggregate_function: str, is_reversed_score, *args, **kwargs):
```
maximilianwerk, line 12: Please add a type hint and a reasonable default for `is_reserved_score`.

rankers/AggregateRanker/__init__.py
florian-hoenicke: Thanks, done

rankers/AggregateRanker/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        self.aggregate_function = aggregate_function
+
+    def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
+        scores = match_idx[self.COL_SCORE]
+        if self.aggregate_function in self.AGGREGATE_FUNCTIONS:
+            aggregated_score = getattr(np, self.aggregate_function)(scores)
+        else:
+            raise ValueError(f'The aggregate function "{self.aggregate_function}" is not defined.')
+        if self.is_reversed_score:
+            aggregated_score = 1. / (1. + aggregated_score)
```
maximilianwerk, line 31: If I am not mistaken, that does not work if we have positive and negative scores mixed. Having a test for this would be great.

While usually there are no negative scores, it can still happen.

rankers/AggregateRanker/tests/test_aggregate_ranker.py
```diff
@@ -0,0 +1,274 @@
+import numpy as np
+
+from jina.executors.rankers import Chunk2DocRanker
+import pytest
+from .. import AggregateRanker
+
+
+def create_data():
```
JoanFM, line 8: can we make it a `fixture`?

JoanFM (commented):
I would vote for:

Changing `Chunk2DocRanker` name to `AggregateRanker` and add the following Rankers:
- SumRanker
- MeanRanker
- ProdRanker
...

rankers/AggregateRanker/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        :param: is_reverse_score: True if a large score of the matched chunks mean a lower score of the matched doc,
+        othewise: False
+
+        """
+        super().__init__(*args, **kwargs)
+        self.is_reversed_score = is_reversed_score
+        if aggregate_function in self.AGGREGATE_FUNCTIONS:
+            self.np_aggregate_function = getattr(np, aggregate_function)
+        else:
+            raise ValueError(f'The aggregate function "{aggregate_function}" is not one of the values: "{self.AGGREGATE_FUNCTIONS}".')
```
maximilianwerk, line 25: We could even be bold and do the following:
```
if hasattr(np, aggregate_function):
    if aggregate_function not in self.AGGREGATE_FUNCTIONS:
        self.logger.warning(f'The following aggregation function is only available for experimental use in jina and not fully tested: {aggregate_function}'
    self.np_aggregate_function = getattr(np, aggregate_functoin)
else:
   raise ...
```

maximilianwerk (changes requested):
Adding tests for negative scores would be great. Actually, even for mixed positive/negative ones.

Apparently, I like consolidating these rankers into one. Less different object to know from a users perspective.

JoanFM:
> Adding tests for negative scores would be great. Actually, even for mixed positive/negative ones.
> 
> Apparently, I like consolidating these rankers into one. Less different object to know from a users perspective.

But where will we put the barrier? because Aggregate is what `Chunk2DocRanker` do right?

maximilianwerk:
> > Adding tests for negative scores would be great. Actually, even for mixed positive/negative ones.
> > Apparently, I like consolidating these rankers into one. Less different object to know from a users perspective.
> 
> But where will we put the barrier? because Aggregate is what `Chunk2DocRanker` do right?

Fair enough. What about
`Chunk2DocRanker` => `AggregationRanker`
The here implemented ranker => `SimpleAggregationRanker`

Then you can still go crazy and build a crazy score computation algorithm by inheriting from the `AggregationRanker` and do more magic. Like a `LambdaMartRanker`.

JoanFM:
> > > Adding tests for negative scores would be great. Actually, even for mixed positive/negative ones.
> > > Apparently, I like consolidating these rankers into one. Less different object to know from a users perspective.
> > 
> > 
> > But where will we put the barrier? because Aggregate is what `Chunk2DocRanker` do right?
> 
> Fair enough. What about
> `Chunk2DocRanker` => `AggregationRanker`
> The here implemented ranker => `SimpleAggregationRanker`
> 
> Then you can still go crazy and build a crazy score computation algorithm by inheriting from the `AggregationRanker` and do more magic. Like a `LambdaMartRanker`.

I like the idea! Then we can rebrand this as `SimpleAggregationRanker` and then apply the Chunk2DocRanker rebranding in different PR? and also remove Min and MaxRanker?

florian-hoenicke:
Sounds good to me!
If everyone else is fine with that solution I, will take care of the Min/MaxRanker pr and the Chunk2DocRanker rebranding pr.

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
```diff
@@ -0,0 +1,274 @@
+import numpy as np
+
```
JoanFM, line 2: would it be possible to leverage `pytest` parametrization for these tests?

rankers/SimpleAggregateRanker/manifest.yml
```diff
@@ -0,0 +1,13 @@
...
+type: pod
+description: :class:`SimpleAggregateRanker` calculates the score of the matched doc from the matched chunks.
+  For each matched doc, the score is aggregated from all the matched chunks belonging to that doc.
+author: Jina AI Dev-Team (dev-team@jina.ai)
+url: https://jina.ai
+vendor: Jina AI Limited
+documentation: https://github.com/jina-ai/jina-hub
+version: 0.0.6
+license: apache-2.0
+keywords: [ranker, toyexample, demo]
```
JoanFM, line 13: toyexample and demo seem like bad `keywords`

rankers/SimpleAggregateRanker/manifest.yml
```diff
@@ -0,0 +1,13 @@
+manifest_version: 1
+name: SimpleAggregateRanker
+kind: ranker
+type: pod
+description: :class:`SimpleAggregateRanker` calculates the score of the matched doc from the matched chunks.
```
JoanFM, line 5: `SimpleAggregateRanker` combines (or reduces?) the score of several documents into one unique score. It is used to aggregate the chunk scores into parent document scores

JoanFM (commented):
What do you like better (Aggregate or Reduce)? @florian-hoenicke @maximilianwerk

maximilianwerk:
Please do not yet remove the `MinRanker` and `MaxRanker`. This can be done in a second step, once we adapt the rest of our codebase to the new ranker. They should rather get a deprecation flag and be removed in another jina version.

rankers/AggregateRanker/tests/test_aggregate_ranker.py
florian-hoenicke: thanks - it removes quite some lines - done

rankers/AggregateRanker/__init__.py
florian-hoenicke: Thanks, I added the type hint. Having a reasonable default for `is_reversed_score` is quite difficult. It depends on the used metric the chunks are compared with. If the chunk score is a distance function `is_reversed_score` should be set to `True` and if the chunk score is a similarity function `is_reversed_score` should be set to `False`.
If the ranker could somehow retrieve the type (similarity or distance) from the flow context, we would not even need that parameter.

rankers/AggregateRanker/__init__.py
florian-hoenicke: If we do this one could use var, std and let's say np.count_nonzero.
On the other hand... maybe I'm a bit paranoid, but to me this would feel like introducing a potenzial security vulnerability. If at the end the aggregate_function string comes from outside, one could call any np function. I don't have a concrete example where it is actually harmful, but if we just allow certain functions, we don't even have to think about that case.

rankers/MaxRanker/__init__.py
```diff
@@ -1,4 +1,7 @@
 from jina.executors.rankers import Chunk2DocRanker
+import warnings
```
JoanFM, line 2: I would put it inside the `__init__` or `post_init` so that is not shown in every `jina` usage?

rankers/SimpleAggregateRanker/manifest.yml
florian-hoenicke: Aggregate, reduce, accumulate or compress are all synonyms in functional programming. So I have no preference. But I wouldn't use `combine`

JoanFM: yes, good point, anyway is good

rankers/MaxRanker/__init__.py
florian-hoenicke: nice catch! done

rankers/SimpleAggregateRanker/manifest.yml
florian-hoenicke: changed it to `aggregates`

rankers/SimpleAggregateRanker/manifest.yml
florian-hoenicke: True, I thought it has to be like this because it was the same in Min/MaxRanker.
I updated the keywords. Thanks.

rankers/AggregateRanker/__init__.py
florian-hoenicke: This function is for converting distance measures into similarity measures. Since distances can not be negative, we are fine - assuming the user is not using it for a similarity measure. To prevent misconfigurations, I think in the long term, we should put this conversion some levels up the call stack where we know what kind of metric was used.

JoanFM: I think making it robust to have no `ZeroDivisionError` (and just have a warning could be good). In jina is desired that `drivers` tend to the best of the efforts not to know what the underlying executors are, this way we can split as much as possible the responsibilities and have different developer experiences.

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
florian-hoenicke: sure, I updated the code to use parametrization.

rankers/AggregateRanker/__init__.py
florian-hoenicke: Okay, in that case, I would like to raise an exception even when the score is negative and `is_reversed_score` is used.
Otherwise around aggregated_score = -1, it will swap between low negative and high positive numbers. I think this will not be used on purpose.
image

florian-hoenicke: done

rankers/SimpleAggregateRanker/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+        if aggregate_function in self.AGGREGATE_FUNCTIONS:
+            self.np_aggregate_function = getattr(np, aggregate_function)
+        else:
+            raise ValueError(f'The aggregate function "{aggregate_function}" is not one of the values: "{self.AGGREGATE_FUNCTIONS}".')
+
+    def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
+        scores = match_idx[self.COL_SCORE]
+        aggregated_score = self.np_aggregate_function(scores)
+        if self.is_reversed_score:
+            if aggregated_score < 0:
```
JoanFM, line 31: I think here we just need to check `DivByZeroError` but raising an exception for something that is well defined should not happen.

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
```diff
@@ -0,0 +1,166 @@
...
+    ),
+    (
+            chunk_scores(),
+            'prod',
+            False,
+            [3, 1, 2],
+            [0.1, 0.2 * 0.4 * 0.5, 0.25 * 0.25 * 0.15 * 0.15]
+    ),
+    (
+            chunk_scores(),
```
JoanFM, line 136: just fyi, if it does not variate `chunk_scores` does not need to be parametrized and can be simply passed as a fixture.

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    return match_idx_numpy, query_chunk_meta, match_chunk_meta
+
+
+def assert_document_order(doc_idx):
+    for i in range(0, len(doc_idx) - 1):
+        assert doc_idx[i][1] > doc_idx[i + 1][1]
+
+
+@pytest.mark.parametrize("chunk_scores,aggregate_function,is_reversed_score,doc_ids,doc_scores", [
```
JoanFM, line 58: single quotes and space between  if possible

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
```diff
@@ -0,0 +1,166 @@
+import numpy as np
+
+from jina.executors.rankers import Chunk2DocRanker
+import pytest
+from .. import SimpleAggregateRanker
+
+
+def chunk_scores(is_positive_score=True):
```
JoanFM, line 8: is the `is_positive_score` ever changed in this test?

JoanFM (changes requested):

rankers/SimpleAggregateRanker/__init__.py
```diff
@@ -28,8 +28,7 @@ def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwa
         scores = match_idx[self.COL_SCORE]
         aggregated_score = self.np_aggregate_function(scores)
         if self.is_reversed_score:
-            if aggregated_score < 0:
-                raise ValueError(f'Setting is_reversed_score to "{self.is_reversed_score}" does not allow negative '
-                                 f'chunk scores (aggregated_score is {aggregated_score}).')
+            if aggregated_score == 0:
```
JoanFM, line 31: what it does not allow is having the `score` to -1, right?

JoanFM (changes requested):

rankers/MinRanker/__init__.py
```diff
@@ -3,11 +3,17 @@
 
 class MinRanker(Chunk2DocRanker):
     """
-    :class:`MinRanker` calculates the score of the matched doc form the matched chunks. For each matched doc, the score
+    :class:`MinRanker` calculates the score of the matched doc from the matched chunks. For each matched doc, the score
```
JoanFM, line 6: need to update the `manifest.yml` version for it to be `tested` and `pushed`. Same for MinRanker

rankers/SimpleAggregateRanker/__init__.py
florian-hoenicke: I made a change accordingly. Now, only zero is forbidden as result of an aggregate function.
However, this will lead to unpredictable behaviour. Let's say we allow `negative distances` in the chunk matches and use the `mean aggregate function.` As a result we would get sometimes positive and sometimes negative doc scores.
So the exception might occur in some rare cases where the mean value actually becomes zero. I would like to avoid this.
Also, the `reversed_score function` is only used for distances and distances can not be smaller than zero.

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
florian-hoenicke: thanks

rankers/SimpleAggregateRanker/__init__.py
JoanFM: Calling with distances is what would happen most of the time but is not ensured. It could be called with any kind of score. I think the only thing we need to avoid is having a DivByZero which does not happen if aggregate score is 0 right?

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
florian-hoenicke: yes I had one test case where it was set to False. 
Now I use the `factor` parameter to define the negative number case (-1), positive case (1) and zero case (0).

rankers/SimpleAggregateRanker/__init__.py
florian-hoenicke: sorry, I fixed it

rankers/SimpleAggregateRanker/tests/test_aggregate_ranker.py
florian-hoenicke: done

rankers/MinRanker/__init__.py
florian-hoenicke: thanks, I updated the version for MinRanker and MaxRanker

rankers/SimpleAggregateRanker/__init__.py
florian-hoenicke: Yes division by zero does not happen anymore. Now it throws an error.

bhavsarpratik:
recheckcla

bhavsarpratik:
Hi @florian-hoenicke for this superb PR. You need to sign CLA  as shown above before we can merge.

florian-hoenicke:

maximilianwerk (approved):

JoanFM (approved):

PR #485, opened 2020-11-09
feat: add Hamming distance evaluator

smy0428:
add Hamming distance evaluator for text


JoanFM:
Thanks for your contribution. Unfortunately, as you didn't set up your local git config correctly, the author of your commit does not match with your Github account. This poses problems:
Your contribution can not be credited, and will not show up in the list of contributors (as bad git config = anonymous)
We can not accept CLA for anonymous contribution.
Please note that this does not only affect your contribution to Jina, but also affects all your contribution to anything on Github. So better fix it, once for all.
Here is what you can do:
```
git config user.name "Correct new name"
git config user.email "correct@new.email" -> match with your Github registered email address
```
Then fix the commit message by
```
git commit --amend --author="Author Name <email@address.com>" --no-edit
git push --force
```

evaluators/text/HammingDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def metric(self):
+        return 'HammingDistance'
+
+    def evaluate(self, actual: str, desired: str):
+        if len(actual) != len(desired):
+            raise ValueError("Undefined for sequences of unequal length.")
```
theUnkownName, line 16: Hi, thanks for your contributions!
We use single quotations, could you change those?

theUnkownName (changes requested):

evaluators/text/HammingDistanceEvaluator/__init__.py
smy0428: of course!

evaluators/text/HammingDistanceEvaluator/tests/test_hammingdistanceevaluator.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from .. import HammingDistanceEvaluator
+
+
+@pytest.mark.parametrize('actual, desired, distance',
+    [
+        ('hello', 'hello', 0.0),
```
JoanFM, line 8: what happens with sentences with space , commas or other punctuations?

JoanFM (changes requested):

evaluators/text/HammingDistanceEvaluator/tests/test_hammingdistanceevaluator.py
smy0428: I think Hamming Distance is supposed to represent the low-level method to compare between two string variables. 
Hence, if we have some input like sentences or other complicated. It is better to use other evaluators to compute the distance.

JoanFM: But since we do not have control on how it will be used, it is better to have a test and know how it will behave

smy0428: I got your point. fixed.

bwanglzu:
@JoanFM @theUnkownName shall we proceed with this PR? looks good now

evaluators/text/HammingDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,22 @@
+from jina.executors.evaluators.text import BaseTextEvaluator
+
+class HammingDistanceEvaluator(BaseTextEvaluator):
+    """A:class:`HammingDistanceEvaluator` Gives the Hamming distance between result and groundtruth string..
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
bwanglzu, line 8: i don't see the point of pouting the constructor here.

JoanFM:
> @JoanFM @theUnkownName shall we proceed with this PR? looks good now

Sure

JoanFM (approved):

PR #486, opened 2020-11-09
feat: add general Minkowski distance evaluator

smy0428:
add Minkowski distance evaluator for more general case p > 0.


evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+        :param desired: the expected embedding of the document
+        :return the evaluation metric value for the request document
+        """
+        actual = expand_vector(actual)
+        desired = expand_vector(desired)
+        order = kwargs["order"]
+        return _minkowski_Distance(actual, desired, order)
+
+
+def _minkowski_Distance(actual, desired, order):
```
JoanFM, line 30: We tend to use underscore case, please do not mix it with camelCase.

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    if order <= 0:
+        raise ValueError("Order must be positive.")
+    minkowskiDistExp = np.sum(np.power(np.abs(actual - desired), order))
```
JoanFM, line 33: Underscore case to be used here

JoanFM (changes requested):

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
smy0428: thanks for all your nice suggestions.

JoanFM:
Hey @smy0428 can you please take a look at your git config? It seems not to be set properly.

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
smy0428: Screen Shot 2020-11-09 at 10 41 57 PM

sorry it is my first time to pr. I have tried to set the username and user email. And I am not sure if it works or not.

JoanFM: Now you need to change your commit messages

smy0428: it turns out that my Git email is another one.

smy0428:
I have reset my git email. Please let me know if everything is fine now.

JoanFM:
> I have reset my git email. Please let me know if everything is fine now.


You need to change the `commits` to have them with your account

smy0428:
I see. I rebased the previous commits.

JoanFM:
recheckcla

JoanFM:
can u do the same for every PR?

smy0428:
I will do it for every commit.

smy0428:
Now they should be done.

JoanFM:
Hey @smy0428 , we need you to sign the `CLA` agreement as stated above

smy0428:

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def evaluate(self, actual: 'np.array', desired: 'np.array', *args, **kwargs) -> float:
+        """"
+        :param actual: the embedding of the document (resulting from an Encoder)
+        :param desired: the expected embedding of the document
+        :return the evaluation metric value for the request document
+        """
+        actual = expand_vector(actual)
+        desired = expand_vector(desired)
+        order = kwargs["order"]
```
JoanFM, line 26: this order will not come here, the `order` must be an `attribute` of the class, but will not come in this `evaluate` `kwargs`. We may need to remove the `variable arguments` from the signatures.

JoanFM (changes requested):

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+        return _minkowski_distance(actual, desired, order)
+
+
+def _minkowski_distance(actual, desired, order):
+    if order <= 0:
+        raise ValueError("Order must be positive.")
```
JoanFM, line 32: please use `single` quote

JoanFM (changes requested):

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
smy0428: Yes, fixed

evaluators/embedding/MinkowskiDistanceEvaluator/__init__.py
smy0428: Are you meaning that the order should be an attribute of the class or should be an attribute of an instance?
My understanding is that this is a general evaluator which accepts any positive p as its order. Therefore it might be an attribute of an instance.

JoanFM: yes, it would be an attribute of an instance sorry

JoanFM: Hey @smy0428 would you please fix this and make it an instance attribute?

JoanFM (changes requested):

smy0428:
Yes, of course. Sorry for the late reply.
Shall I also change the parameters of the constructor function somehow? I am not sure how I should include it as an instance attribute

JoanFM:
> Yes, of course. Sorry for the late reply.
> Shall I also change the parameters of the constructor function somehow? I am not sure how I should include it as an instance attribute

Yes the parameter for the constructor is the way to go.

JoanFM:
@bhavsarpratik this is the executor we discussed. Do you propose another naming?

smy0428:
I have implemented a new ctor

bhavsarpratik:
@JoanFM I am ok with the current

JoanFM (approved):

PR #489, opened 2020-11-12
feat: add VSETextEncoder from cxmodal

JoanFM:


encoders/nlp/VSETextEncoder/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, path: str = 'runs/f30k_vse++_vggfull/model_best.pth.tar',
+                 vocab_path: str = 'vocab/f30k_vocab.pkl', *args, **kwargs):
+        """
+        :path : path where to find the model.pth file
+        :vocab_path : path where to find the vocab.pkl file
+        """
+        super().__init__(*args, **kwargs)
+        self.path = path
+        self.vocab_path = vocab_path
```
nan-wang, line 36: Where the model file is stored?

JoanFM: It is downloaded at docker build time. However, it is too big (because the author put almost every single experiment inside the same .tar) I need a place where I can publicly host a 600MB file, wherr could I do that?

JoanFM: Same goes for the vocabulary

JoanFM: Ah no, it is faster here in github, I think is good to be merged @nan-wang !

We are not uploading any heavy file inside the repo

nan-wang (approved):
LGTM👍

PR #753, opened 2020-11-19
fix: change default transformer model to distilbert

bhavsarpratik:
As per this the distil version is smaller yet better
https://github.com/UKPLab/sentence-transformers#performance


JoanFM (commented):
Why do we put by default the distilled version? I would say the current one seems like a better default option

bhavsarpratik:
As per this the distil version is smaller yet better
https://github.com/UKPLab/sentence-transformers#performance

JoanFM:
> As per this the distil version is smaller yet better
> https://github.com/UKPLab/sentence-transformers#performance

So can you add this in the description of the PR and change also in the torch version?

Also will it potentially break the examples tests?

bhavsarpratik:
In this PR, I have changed for both TF and torch.
We can keep the examples intact by specifying the current model `bert-base-uncased` and pooling `auto` in the encoder pod if it is using the default.

bhavsarpratik:
Southpark and multires have model name and pooling defined in encode pod. So they will not break.

JoanFM:
Hey @bhavsarpratik can you please rebase?

encoders/nlp/TransformerTFEncoder/tests/test_transformertfencoder.py
```diff
@@ -28,27 +28,23 @@ def test_metas(tmpdir, random_workspace_name):
 
 encoders_parameters = [
     {
-        "pretrained_model_name_or_path": 'sentence-transformers/distilbert-base-nli-stsb-mean-tokens',
-        "model_save_path": 'distilbert-base-nli-stsb-mean-tokens',
-    },
-    {
-        "pretrained_model_name_or_path": 'bert-base-uncased',
-        "model_save_path": 'bert-base-uncased',
+        "pretrained_model_name_or_path": 'distilbert-base-uncased',
```
JoanFM, line 31: would it be good to maybe test more models that are popular? and maybe build a matrix of parameters by using `pytest parametrize` to have kind of a cartesian product of different parameters for extensive testing? Like cover 3 or 4 popular models with each of them will different pooling strategy and even layer index?

JoanFM, line 31: would it be good to maybe test more models that are popular? and maybe build a matrix of parameters by using `pytest parametrize` to have kind of a cartesian product of different parameters for extensive testing? Like cover 3 or 4 popular models with each of them will different pooling strategy and even layer index?

bhavsarpratik: Indeed! Let's do it in another PR

JoanFM: ok, add an issue with the idea

JoanFM (approved):

PR #1135, opened 2020-11-24
feat: add tirg image encoder

bhavsarpratik:


bhavsarpratik:
@bwanglzu @JoanFM emb mismatch is back

encoders/image/TirgImageEncoder/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+import os
+import sys
+import pickle
+
+from jina.executors.decorators import batching, as_ndarray
+from jina.executors.encoders.frameworks import BaseTorchEncoder
+from jina.excepts import PretrainedModelFileDoesNotExist
+
+sys.path.append(".")
+from img_text_composition_models import TIRG
```
JoanFM, line 13: Hide these import inside functions, and I think we can skip the `sys path` (u can see VSEImageEncoder for reference)

encoders/image/TirgImageEncoder/img_text_composition_models.py
```diff
@@ -0,0 +1,259 @@
...
+# limitations under the License.
+# ==============================================================================
+
+"""Models for Text and Image Composition."""
+
+import numpy as np
+import torch
+import torchvision
+import torch.nn.functional as F
+# from jina.hub.encoders.image.TirgEncoder import text_model
```
JoanFM, line 25: remember to remove this here

encoders/image/TirgImageEncoder/text_model.py
```diff
@@ -0,0 +1,126 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
```
JoanFM, line 1: Is it possible to remove this file? Or it fails?

bwanglzu (dismissed):
lgtm!

JoanFM:
@bhavsarpratik can u try the following?

Can u copy this into jina subfolder `jina/hub ...` and without installing these dependencies run `jina check` . See if it complains about this Executor not being properly loaded?

bhavsarpratik:
@JoanFM It's not complaining

encoders/image/TirgImageEncoder/text_model.py
bhavsarpratik: It is needed :(

encoders/image/TirgImageEncoder/__init__.py
```diff
@@ -9,6 +9,8 @@
 from jina.executors.encoders.frameworks import BaseTorchEncoder
 from jina.excepts import PretrainedModelFileDoesNotExist
 
+from .img_text_composition_models import TIRG
```
JoanFM, line 12: Hide this inside function or hide the imports in the file in subfunctions. Otherwise the ugly warning will appear

JoanFM (changes requested):

encoders/image/TirgImageEncoder/__init__.py
JoanFM: Jina inspects every module when starts and will hit here and import this one and may find that torch is not imported and start complaining

bhavsarpratik: What you say is true. When I moved it inside the function, it is not able to import `img_text_composition_models`. Why does VSE has it outside at the top?

JoanFM: because I hid the external packages in the `module.py` and so. Ok, I think it is explored in anyway because of the `py_modules`? So maybe we need to hide them in those external files

bhavsarpratik: That is doable but what about the current fail message
`could not determine a constructor for the tag '!TirgImageEncoder`

JoanFM: Really weird, have u run locally the `jina hub build --prune-images --raise-error --host-info --test-uses --daemon .` from the folder inside this hub?

bhavsarpratik: I haven't run it locally as I thought this will work but now it fails. Docker build takes a lot of time on local due to torch download. But I have a workaround for it by adding `pip install torchvision` at the start of dockerfile.

JoanFM: does it fail with the same message as here?

JoanFM:
I think there is a problem with our `builder`

JoanFM (approved):

PR #1180, opened 2020-11-26
ci: add updater and make builder push by default

cristianmtr:
+ Add new Hub Updater action
+ Make `ci.yml` hub builder also push images


rutujasurve94 (changes requested):
LGTM

rutujasurve94 (approved):

PR #1232, opened 2020-12-03
feat: add update and delete to mongodbindexer

cristianmtr:
Add update and delete to MongoDBIndexer

### DIFF
--- a/indexers/keyvalue/MongoDBIndexer/__init__.py
+++ b/indexers/keyvalue/MongoDBIndexer/__init__.py
@@ -1,9 +1,7 @@
 from typing import Optional, Iterator
 
-from jina.proto import jina_pb2
-from jina.logging.logger import JinaLogger
-from jina.helper import cached_property
 from jina.executors.indexers.keyvalue import BinaryPbIndexer
+from jina.helper import cached_property
 
 
 class MongoDBIndexer(BinaryPbIndexer):
@@ -46,13 +44,15 @@ class MongoDBIndexer(BinaryPbIndexer):
     def get_query_handler(self):
         return self.handler
     
-    def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
-        mongo_docs = [{'_id': i, 'values': j} for i, j in zip(keys, values)]
-        
+    def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs) -> None:
+        total_inserted_ids = []
         with self.write_handler as mongo_handler:
-            inserted_ids = mongo_handler.insert(documents=mongo_docs)
+            for i, j in zip(keys, values):
+                doc = {'_id': i, 'values': j}
+                inserted_ids = mongo_handler.insert(documents=[doc])
+                total_inserted_ids.extend(inserted_ids)
         
-        if inserted_ids and len(inserted_ids) != len(keys):
+        if total_inserted_ids and len(total_inserted_ids) != len(list(keys)):
             self.logger.error(f'Mismatch in mongo insert')
     
     @cached_property
@@ -61,7 +61,16 @@ class MongoDBIndexer(BinaryPbIndexer):
 
     def query(self, key: int, *args, **kwargs) -> Optional[bytes]:
         with self.query_handler as mongo_handler:
-            result = mongo_handler.find({'_id': key})
+            result = mongo_handler.find(key)
         
         if result:
             return result
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs) -> None:
+        """NOTE: this completely replaces the document with the key"""
+        with self.query_handler as mongo_handler:
+            mongo_handler.update(keys, values)
+
+    def delete(self, keys: Iterator[int], *args, **kwargs) -> None:
+        with self.query_handler as mongo_handler:
+            mongo_handler.delete(keys)
--- a/indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
+++ b/indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Union, Optional, Iterator
+from typing import Dict, Optional, Iterator
 
 from jina.logging.logger import JinaLogger
 
@@ -13,10 +13,11 @@ class MongoDBHandler:
     MongoDB has no access control by default, hence can be used without username:password.
     If username & password are passed, we need to create it (can be changed to existing un:pw)
     """
-    def __init__(self, 
-                 hostname: str = '127.0.0.1', 
-                 port: int = 27017, 
-                 username: str = None, 
+
+    def __init__(self,
+                 hostname: str = '127.0.0.1',
+                 port: int = 27017,
+                 username: str = None,
                  password: str = None,
                  database: str = 'defaultdb',
                  collection: str = 'defaultcol'):
@@ -33,10 +34,10 @@ class MongoDBHandler:
         else:
             self.connection_string = \
                 f'mongodb://{self.hostname}:{self.port}'
-        
+
     def __enter__(self):
         return self.connect()
-    
+
     def connect(self) -> 'MongoDBHandler':
         import pymongo
         try:
@@ -52,27 +53,31 @@ class MongoDBHandler:
         except Exception as exp:
             raise MongoDBException(exp)
         return self
-        
+
     @property
     def database(self):
         return self.client[self.database_name]
-    
+
     @property
     def collection(self):
         return self.database[self.collection_name]
-    
-    def find(self, query: Dict[str, Union[Dict, List]]) -> None:
+
+    def find(self, key: int) -> Optional[bytes]:
         import pymongo
         try:
-            return self.collection.find(query)
+            cursor = self.collection.find({'_id': key})
+            cursor_contents = list(cursor)
+            if cursor_contents:
+                return cursor_contents[0]
+            return None
         except pymongo.errors.PyMongoError as exp:
             self.logger.error(f'Got an error while finding a document in the db {exp}')
-        
+
     def insert(self, documents: Iterator[Dict]) -> Optional[str]:
         import pymongo
         try:
             result = self.collection.insert_many(documents)
-            self.logger.debug(f'inserted documents in the database')
+            self.logger.debug(f'inserted {len(result.inserted_ids)} documents in the database')
             return result.inserted_ids
         except pymongo.errors.PyMongoError as exp:
             self.logger.error(f'got an error while inserting a document in the db {exp}')
@@ -83,3 +88,30 @@ class MongoDBHandler:
             self.client.close()
         except pymongo.errors.PyMongoError as exp:
             raise MongoDBException(exp)
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        import pymongo
+        try:
+            count = self.collection.delete_many({'_id': {'$in': list(keys)}}).deleted_count
+            self.logger.debug(f'deleted {count} documents in the database')
+        except pymongo.errors.PyMongoError as exp:
+            self.logger.error(f'got an error while deleting a document in the db {exp}')
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        import pymongo
+        try:
+            # update_many updates several keys with the same op. / data.
+            # we need this instead
+            count = 0
+            for k, new_doc in zip(keys, values):
+                new_doc = {'_id': k, 'values': new_doc}
+                inserted_doc = self.collection.find_one_and_replace(
+                    {'_id': k},
+                    new_doc
+                )
+                if inserted_doc == new_doc:
+                    count += 1
+            self.logger.debug(f'updated {count} documents in the database')
+            return
+        except pymongo.errors.PyMongoError as exp:
+            self.logger.error(f'got an error while updating documents in the db {exp}')
--- a/indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
+++ b/indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
@@ -1,62 +1,77 @@
-import string
 import random
+from collections import Iterator
 
 import numpy as np
-from google.protobuf.json_format import MessageToJson
+from jina import Document
 
-from jina.drivers.helper import array2pb
-from jina.proto import jina_pb2, uid
 from .. import MongoDBIndexer
 
 
 def random_docs(num_docs, chunks_per_doc=5, embed_dim=10, jitter=1):
     c_id = 3 * num_docs  # avoid collision with docs
     for j in range(num_docs):
-        d = jina_pb2.Document()
-        d.tags['id'] = j
-        d.text = b'hello world doc id %d' % j
-        d.embedding.CopyFrom(array2pb(np.random.random([embed_dim + np.random.randint(0, jitter)])))
-        d.id = uid.new_doc_id(d)
-        for k in range(chunks_per_doc):
-            c = d.chunks.add()
-            c.text = 'i\'m chunk %d from doc %d' % (c_id, j)
-            c.embedding.CopyFrom(array2pb(np.random.random([embed_dim + np.random.randint(0, jitter)])))
-            c.tags['id'] = c_id
-            c.tags['parent_id'] = j
-            c_id += 1
-            c.parent_id = d.id
-            c.id = uid.new_doc_id(c)
+        with Document() as d:
+            d.tags['id'] = j
+            d.text = b'hello world doc id %d' % j
+            d.embedding = np.random.random([embed_dim + np.random.randint(0, jitter)])
+            for k in range(chunks_per_doc):
+                with Document() as c:
+                    c.text = 'i\'m chunk %d from doc %d' % (c_id, j)
+                    c.embedding = np.random.random([embed_dim + np.random.randint(0, jitter)])
+                    c.tags['id'] = c_id
+                    c.tags['parent_id'] = j
+                    c_id += 1
+                    c.parent_id = d.id
+                d.chunks.add(c)
         yield d
 
 
 def test_mongodbindexer():
-    keys = []
-    values = []
-    
     num_docs = 5
-    docs = list(random_docs(num_docs=num_docs, 
+    docs = list(random_docs(num_docs=num_docs,
                             chunks_per_doc=3))
-    keys = [uid.id2hash(doc.id) for doc in docs]
+    keys: Iterator[int] = iter([doc.id for doc in docs])
     values = [doc.SerializeToString() for doc in docs]
-    
+
     query_index = random.randint(0, num_docs - 1)
-    query_id = docs[query_index].id
-    query_key = uid.id2hash(query_id)
+    # adding type annotations to remove further warnings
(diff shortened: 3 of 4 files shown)


indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +83,19 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.client.close()
         except pymongo.errors.PyMongoError as exp:
             raise MongoDBException(exp)
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        import pymongo
+        try:
+            count = 0
+            for k in keys:
+                result = self.collection.delete_one({'_id': k})
```
JoanFM, line 92: is there a batch delete?

If we count the number of deletes in this way is because expect it to be different than the amount of deletes we expect right? Then the relevant info to debug is the difference or the ones that you did not manage to delete (or the keys that u did in fact delete)

cristianmtr: Yep, found the `delete_many`. Needed a special operator `$in`

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -52,28 +53,32 @@ def connect(self) -> 'MongoDBHandler':
...
-        
+
     def insert(self, documents: Iterator[Dict]) -> Optional[str]:
         import pymongo
         try:
             result = self.collection.insert_many(documents)
-            self.logger.debug(f'inserted documents in the database')
-            return result.inserted_ids
+            self.logger.debug(f'inserted {len(result.inserted_ids)} documents in the database')
+            return
```
JoanFM, line 81: no need to add return in this case

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +88,31 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.client.close()
         except pymongo.errors.PyMongoError as exp:
             raise MongoDBException(exp)
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        import pymongo
+        try:
+            count = self.collection.delete_many({'_id': {'$in': keys}}).deleted_count
+            self.logger.debug(f'deleted {count} documents in the database')
+            return
```
JoanFM, line 97: same here, I am not sure how it would play if we ever add a `finally`

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +88,31 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            # we need this instead
+            count = 0
+            for k, new_doc in zip(keys, mongo_docs):
+                inserted_doc = self.collection.find_one_and_replace(
+                    {'_id': k},
+                    new_doc
+                )
+                if inserted_doc == new_doc:
+                    count += 1
+            self.logger.debug(f'updatd {count} documents in the database')
```
JoanFM, line 115: typo in `updatd`.

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
```diff
@@ -1,61 +1,85 @@
...
+    # adding type annotations to remove further warnings
+    keys: Iterator[int] = [doc.id for doc in docs]
     values = [doc.SerializeToString() for doc in docs]
-    
+
     query_index = random.randint(0, num_docs - 1)
-    query_id = docs[query_index].id
-    query_key = uid.id2hash(query_id)
+    # adding type annotations to remove further warnings
+    query_key: int = docs[query_index].id
```
JoanFM, line 39: why is this annotation needed?

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
```diff
@@ -1,61 +1,85 @@
...
+        count = 0
+        for key in keys:
+            result = mongo_query.query(key)
+            if result:
+                count += 1
+
+        assert count == 0
 
 
 if __name__ == "__main__":
```
JoanFM, line 85: can u remove this `main`?

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
cristianmtr: `.query`, `.add` etc. expect `key` to be `int`, but `doc.id` is `UniqueId`. It's a warning from PyCharm, no problems at run time. Not sure why

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
cristianmtr: Done

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
cristianmtr: Done

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
cristianmtr: Done

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +87,30 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def delete(self, keys: Iterator[int], *args, **kwargs):
+        import pymongo
+        try:
+            count = self.collection.delete_many({'_id': {'$in': keys}}).deleted_count
+            self.logger.debug(f'deleted {count} documents in the database')
+        except pymongo.errors.PyMongoError as exp:
+            self.logger.error(f'got an error while deleting a document in the db {exp}')
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], ):
```
JoanFM, line 99: add *args and **kwargs as per discussion here jina-ai/jina#1285

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
```diff
@@ -1,62 +1,82 @@
...
-
-if __name__ == "__main__":
-    test_mongodbindexer()
+        mongo_query.delete(keys)
+
+    with MongoDBIndexer() as mongo_query:
+        count = 0
+        for key in keys:
+            result = mongo_query.query(key)
+            if result:
```
JoanFM, line 79: I would assert `not result` as well then right?

JoanFM (changes requested):

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
cristianmtr: Done

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
cristianmtr: `assert result is None`, but yes

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
```diff
@@ -1,62 +1,78 @@
...
+
+    with MongoDBIndexer() as mongo_query:
+        result = mongo_query.query(key=query_key)
+        assert result['_id'] == query_key
+        d = Document()
+        d.ParseFromString(result['values'])
+        assert d.text == query_text
+
+    # update
+    new_docs = list(random_docs(num_docs=num_docs, chunks_per_doc=3))
```
maximilianwerk, line 54: Not sure, but that feels like it introduces id collisions. We should switch from explicit `id` definitions to random ids, as soon as it is implemented here: https://github.com/jina-ai/jina/pull/1392

Anyhow, that does not block this PR.

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +87,30 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        import pymongo
+        mongo_docs = [{'_id': i, 'values': j} for i, j in zip(keys, values)]
+        try:
+            # update_many updates several keys with the same op. / data.
+            # we need this instead
+            count = 0
+            for k, new_doc in zip(keys, mongo_docs):
+                inserted_doc = self.collection.find_one_and_replace(
```
maximilianwerk, line 107: Replace sounds like MongoDB will overwrite the old content instead of updating it. Is that true? Then we first need to get the document from MongoDB, update it and only then replace the old content.

cristianmtr: What do you mean? Don't we just replace the document with the respective `id` with the values of the `new_doc`? 

For update we need to use mongo update operators, like these https://docs.mongodb.com/manual/reference/operator/update/#id1. We then need to specify the specific fields we're updating (if we use `$set`). But we don't get that sort of request from the `update` method, right? ` values: Iterator[bytes]` represents an entire new document value, not a more granular field update

JoanFM: not for now but it will be included in the future

florian-hoenicke (changes requested):

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
cristianmtr: Hm but is it then within the scope of this PR to handle that? Cause it will also require to establish how to read the `bytes` into JSON

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -83,3 +87,30 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        import pymongo
+        try:
+            count = self.collection.delete_many({'_id': {'$in': keys}}).deleted_count
+            self.logger.debug(f'deleted {count} documents in the database')
+        except pymongo.errors.PyMongoError as exp:
+            self.logger.error(f'got an error while deleting a document in the db {exp}')
+
+    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+        import pymongo
+        mongo_docs = [{'_id': i, 'values': j} for i, j in zip(keys, values)]
```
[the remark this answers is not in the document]
cristianmtr: Ah yes, I'll rewrite it to be an iterator

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
```diff
@@ -1,62 +1,78 @@
...
-    keys = []
-    values = []
-    
     num_docs = 5
-    docs = list(random_docs(num_docs=num_docs, 
+    docs = list(random_docs(num_docs=num_docs,
                             chunks_per_doc=3))
-    keys = [uid.id2hash(doc.id) for doc in docs]
+    # adding type annotations to remove further warnings
+    keys: Iterator[int] = [doc.id for doc in docs]
```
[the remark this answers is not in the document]
cristianmtr: Yes, but then I get the PyCharm warning of "Expected type 'Iterator[int]', got 'Iterator[UniqueId]' instead" at `keys: Iterator[int] = iter([doc.id for doc in docs])`

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
maximilianwerk: OK, sorry my misunderstanding, I thought update is only a partial update and not a full overwrite. This should be well documented, since it is non-obvious.

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
```diff
@@ -52,28 +53,31 @@ def connect(self) -> 'MongoDBHandler':
...
+                return cursor_contents[0]
+            return None
         except pymongo.errors.PyMongoError as exp:
             self.logger.error(f'Got an error while finding a document in the db {exp}')
-        
+
     def insert(self, documents: Iterator[Dict]) -> Optional[str]:
         import pymongo
         try:
             result = self.collection.insert_many(documents)
```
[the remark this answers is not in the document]
cristianmtr: I re-wrote this

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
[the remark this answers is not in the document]
cristianmtr: I re-wrote this

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
cristianmtr: I added a note to the docstring. I will also collect all these CRUD gotchas in the Jira ticket

maximilianwerk (commented):
We have quite a different way of doing tests here vs. Redis/LevelDB. Are there specific reasons for it? If no, could we do the same testing approach for both? Or even unify these tests for any Indexer and just call a test function with the specific indexer? Finally, they all should serve the same use-case.

If so, it would be great, to build up this test-case in the core-repo and use it there for the BinaryPBIndexer and then in all other KVIndexer in the hub, non?

If it is not possible to unify it, don't spend too much time on it. But it would be a cool pattern to just write a new indexer and get the test right away with a one-liner.

cristianmtr:
> We have quite a different way of doing tests here vs. Redis/LevelDB. Are there specific reasons for it? If no, could we do the same testing approach for both? Or even unify these tests for any Indexer and just call a test function with the specific indexer? Finally, they all should serve the same use-case.
> 
> If so, it would be great, to build up this test-case in the core-repo and use it there for the BinaryPBIndexer and then in all other KVIndexer in the hub, non?
> 
> If it is not possible to unify it, don't spend too much time on it. But it would be a cool pattern to just write a new indexer and get the test right away with only the custom set-up code.
> 
> @florian-hoenicke @cristianmtr

I agree that it would be nice to have them all look the same. The only reason I did it this way for MongoDBIndexer is because that's how we started doing it for the BinaryPbIndexer.

Also there might be specific indexers that won't pass some tests. E.g. BinaryPbIndexer doesn't support `add` and `update` in the same context. And possibly other indexer-specific issues that haven't been ironed out. 

I think @florian-hoenicke suggested something like this some days ago, but he ran into some of these problems there. Do you remember what those were, or am I misremembering?

florian-hoenicke:
@cristianmtr Yes, I asked if we should write tests for baseclasses and run them on the derived classes. However, as @JoanFM  pointed out, we do not write core tests for the hub.
However, @hanxiao  suggested to add this functionality in the cookie cutter.
I created a ticket accordingly https://github.com/jina-ai/cookiecutter-jina/issues/39
fyi @maximilianwerk

florian-hoenicke (approved):

florian-hoenicke:
@JoanFM can the mongodbindexer be merged?

indexers/keyvalue/MongoDBIndexer/__init__.py
```diff
@@ -46,13 +44,15 @@ def get_create_handler(self):
...
-    def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
-        mongo_docs = [{'_id': i, 'values': j} for i, j in zip(keys, values)]
-        
+    def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs) -> None:
+        total_inserted_ids = []
         with self.write_handler as mongo_handler:
-            inserted_ids = mongo_handler.insert(documents=mongo_docs)
+            for i, j in zip(keys, values):
+                doc = {'_id': i, 'values': j}
+                inserted_ids = mongo_handler.insert(documents=[doc])
```
JoanFM, line 52: why aren't we inserting in batch? it seems mongo_handler accepts more than one document to be inserted at once right?

indexers/keyvalue/MongoDBIndexer/__init__.py
```diff
@@ -46,13 +44,15 @@ def get_create_handler(self):
...
+                total_inserted_ids.extend(inserted_ids)
         
-        if inserted_ids and len(inserted_ids) != len(keys):
+        if total_inserted_ids and len(total_inserted_ids) != len(list(keys)):
```
JoanFM, line 55: why is it needed the transform to `len(list(keys))`?

cristianmtr: `Expected type 'Sized', got 'Iterator[int]' instead`. Just to calm down PyCharm

indexers/keyvalue/MongoDBIndexer/__init__.py
cristianmtr: https://github.com/jina-ai/jina-hub/pull/1232#discussion_r537803852

cristianmtr: Smaller memory footprint

cristianmtr: at the cost of possibly more network IO

JoanFM:
> We have quite a different way of doing tests here vs. Redis/LevelDB. Are there specific reasons for it? If no, could we do the same testing approach for both? Or even unify these tests for any Indexer and just call a test function with the specific indexer? Finally, they all should serve the same use-case.
> 
> If so, it would be great, to build up this test-case in the core-repo and use it there for the BinaryPBIndexer and then in all other KVIndexer in the hub, non?
> 
> If it is not possible to unify it, don't spend too much time on it. But it would be a cool pattern to just write a new indexer and get the test right away with only the custom set-up code.
> 
> @florian-hoenicke @cristianmtr
> 
> EDIT:
> 
> Haha, just saw [jina-ai/cookiecutter-jina#39](https://github.com/jina-ai/cookiecutter-jina/issues/39)
> Seems like the way to go :)

It is intentional not to have Hub executors import anything from the core, so I would avoid that. Also do not share code between executors, it is intended to have `Hub` execturos be independent at the expense of having more code duplication

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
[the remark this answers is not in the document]
JoanFM: The batching will be handled by the `driver`, where have u seen the memory bottleneck? How many documents?

Rather use `batching` definitely.

indexers/keyvalue/MongoDBIndexer/__init__.py
JoanFM: I would definitely go for `batching`

JoanFM: there is some `batching_multi_input` that could serve this purpose

indexers/keyvalue/MongoDBIndexer/mongodbhandler.py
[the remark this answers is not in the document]
cristianmtr: So it seems the `batching_multi_input` doesn't support being used in this way. I will stick to inserting one at a time

JoanFM (approved):

PR #1281, opened 2020-12-10
refactor: remove mean reciprocal rank evaluator

JoanFM:
**Changes introduced**
MeanReciprocalRankEvaluation is handled the accumulation of results from `ReciprocalRankEvaluator` which is the one evaluating document per document


deepankarm:
We still don't have a mechanism to remove a Pod from docker-hub & MongoDB. So even though the code is deleted from here, the Pod exists.

florian-hoenicke:
@JoanFM I'm not so familiar with the evaluation. Could you quickly describe why the MeanReciprocalRankEvaluator has to be deleted?

JoanFM:
> @JoanFM I'm not so familiar with the evaluation. Could you quickly describe why the MeanReciprocalRankEvaluator has to be deleted?

hey @florian-hoenicke ; the point is that evaluation works at doc2doc level; and mean or average computation must be computed by the client to avoid having state. and since we already have ReciprocalRank this is not needed. And actually this will not work with any driver

JoanFM:
> We still don't have a mechanism to remove a Pod from docker-hub & MongoDB. So even though the code is deleted from here, the Pod exists.

at least the code wont be there

florian-hoenicke:
@JoanFM  Thanks. Okay makes totally sense to avoid comparing chunks for evaluation.

florian-hoenicke (approved):

JoanFM:
> @JoanFM Thanks. Okay makes totally sense to avoid comparing chunks for evaluation.

just for reference; is not about chunks. It is about MeanReciprocalRank depending on the evaluation of different queries.

Right now; evaluation works in every pair of (doc, groundtruth) independently from any other document or groundtruth. Like this we mantain independence and can keep our evaluation pods stateless.

MeanReciprocalRank should not be there for the same reason no MeanNDCGEvaluation or MeanPrecision is there, to keep independence between documentation evaluations. Think that an Evaluation Flow can never know for how long u want to accumulate results for average, mean or any other metric. That is for now left at the client consideration.

PR #1285, opened 2020-12-12
fix: update transformer encoders for v4 of transformers library

tadej-redstone:
I have updated the transformer encoders to be compatible with the v4.0 of the transformers library. 

Two main changes:
- instead of passing in an `ndarray` of strings, this is converted to a list - otherwise an error is raised in version 4.0.0 of transformers.
- I have replaced the use of the deprecated `pad_to_max_length` parameter with `padding` parameter. This has no effect whatsoever on behavior of the function.

All these changes do not in any way change the behavior for earlier versions of transformers.

I have also updated the requirements files slightly: for TF, I have replaced `==` with `>=` for tensorflow version - this is consistent with the requirements file for torch transformers. There were no breaking changes in later versions of tensorflow (I have tested locally on 2.3.0), so I don't see the need for this limitation. Also, I have updated the requirement for transformers from `3.0.0` to `3.0.2` - `3.0.2` introduced important bugfixes for tokenizers for bugs in `3.0.0`, while there were no breaking changes.

### DIFF
--- a/encoders/nlp/TransformerTFEncoder/__init__.py
+++ b/encoders/nlp/TransformerTFEncoder/__init__.py
@@ -72,6 +72,8 @@ class TransformerTFEncoder(TFDevice, BaseEncoder):
         self.truncation_strategy = truncation_strategy
         self.model_save_path = model_save_path
 
+        self._padding_strategy = 'max_length' if self.max_length else 'longest'
+
     def __getstate__(self):
         if self.model_save_path:
             if not os.path.exists(self.model_abspath):
@@ -128,15 +130,15 @@ class TransformerTFEncoder(TFDevice, BaseEncoder):
                 self.tokenizer.pad_token = self.tokenizer.eos_token
                 if self.tokenizer.pad_token is None:
                     self.tokenizer.add_special_tokens({'pad_token': '[PAD]'})
-            ids_info = self.tokenizer.batch_encode_plus(data,
+            ids_info = self.tokenizer.batch_encode_plus(list(data),
                                                         max_length=self.max_length,
                                                         truncation=self.truncation_strategy,
-                                                        pad_to_max_length=True)
+                                                        padding=self._padding_strategy)
         except ValueError:
             self.model.resize_token_embeddings(len(self.tokenizer))
-            ids_info = self.tokenizer.batch_encode_plus(data,
+            ids_info = self.tokenizer.batch_encode_plus(list(data),
                                                         max_length=self.max_length,
-                                                        pad_to_max_length=True)
+                                                        padding=self._padding_strategy)
         token_ids_batch = self.array2tensor(ids_info['input_ids'])
         mask_ids_batch = self.array2tensor(ids_info['attention_mask'])
         with self.no_gradients():
--- a/encoders/nlp/TransformerTFEncoder/manifest.yml
+++ b/encoders/nlp/TransformerTFEncoder/manifest.yml
@@ -7,7 +7,7 @@ author: Jina AI Dev-Team (dev-team@jina.ai)
 url: https://jina.ai
 vendor: Jina AI Limited
 documentation: https://github.com/jina-ai/jina-hub
-version: 0.0.7
+version: 0.0.8
 license: apache-2.0
 keywords: [huggingface, transformers, nlp, BERT, tensorflow]
 type: pod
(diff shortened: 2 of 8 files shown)


JoanFM:
Hey @tadej-redstone;

Thanks for your contribution, can you point where there is some documentation indicating that transforming array to list is now the way to go?

Does it come at no performance cost?

Thanks

tadej-redstone:
@JoanFM I will try to find the documentation - but in the meanwhile you can try it out yourself - just remove the `list()` statement, and with transformers 4.0.0 the all the tests will crash. 

I believe there should be no real performance cost, because we are transforming an array of strings (for which I believe numpy has no special optimization) into a list. The whole thing is batched anyway, so we are dealing with lists of length 8 :) If you have any specific concerns, let me know how we could benchmark this.

tadej-redstone:
@JoanFM here's the reason only list is now accepted: there are two types of tokenizers in transformers, slow and fast. In 4.0.0, the default changed from slow to fast tokenizers (see the first breaking change [here](https://github.com/huggingface/transformers/releases/tag/v4.0.0)).

Now fast tokenizers [check that inputs are a list](https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils_fast.py#L361-L364), while slow tokenizers [just try to loop over the input](https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils.py#L520). So that's why passing an array worked before, but does not work now. 

What could be done instead is to force the AutoTokenizer to use the slow tokenisers - but I think this is not a good idea, as the fast tokenizers are preferred in transformers for a reason.

JoanFM:
recheckcla

tadej-redstone:

encoders/nlp/TransformerTFEncoder/__init__.py
```diff
@@ -72,6 +72,8 @@ def __init__(
         self.truncation_strategy = truncation_strategy
         self.model_save_path = model_save_path
 
+        self._padding_strategy = 'max_length' if self.max_length else 'longest'
```
JoanFM, line 75: I know it is now missing, but could a test be added with a max_length set so that this parameter is tested?

JoanFM (changes requested):

encoders/nlp/TransformerTFEncoder/__init__.py
tadej-redstone: @JoanFM  It's good that you mentioned this. If I do this, `test_save_and_load_config` test fails - because nothing gets saved in `save_config()`, so `assert encoder_loaded.max_length == encoder.max_length` fails.


When I say nothing, I mean no attributes specific to `TransformerTorchEncoder`, only params from `BaseExecutor`. The reason that tests were passing so far was that the model was then initiated with default args, and only `max_length` was checked - which was conveniently equal to default `None` in all test examples.

So I think the problem might be that `BaseExecutor` is not saving other arguments at all, or that it offers an interface to save them, but the subclasses are not using it.

So either we investigate this issue first (although I think this is out of scope for this PR) and merge once this is fixed, or we ignore for now, merge, and add the extra test once the issue with saving config is fixed.

JoanFM: could you please add a test where the problem appears?

tadej-redstone: done

JoanFM: it would be cool also to test that 2 sentences that start the same but differ in some words at the end have the same encoding when using a shorter max_length for padding?

tadej-redstone: @JoanFM nevermind what I said before - the tests were failing locally as I was using version 0.8.3 of jina. Updated and now they pass (as they do on CI).

And sure, I'll add such a test

tadej-redstone: @JoanFM yes, I'll add this test.

Also, tests seem to be passing on CI, but they crash locally. Is it because the version in manifest should be updated on each commit for the tests to run, or is it something else?

JoanFM: Is it possible that locally there are some files from previous tests that u did not erase? tests passed properly, no need to raise per each commit

tadej-redstone: I don't know - it's all saved to random directories (tmpdir), so I wouldn't expect this to be a problem. But anyways, CI seems to pass. I have now added the test :)

JoanFM (approved):
Thank you very much for this great contribution!

PR #1286, opened 2020-12-12
feat: add layer_index param to transformer encoders

### LINKED ISSUE #1284: Support outputs from different layers in TransformerTorchEncoder
**Describe the feature**

It would be very useful if it was possible to use other layers, not just the last one, as encoding in `TransformerTorchEncoder`. In a lot of my use cases I find the outputs from the second to last layer more useful than the outputs from the last layer.

**Your proposal**

I suppose an optional argument like `layer_index` could be passed on init, with possible values `-1` (current, default), `-2`, .... In the original BERT paper they also considered summing up embeddings over different layers or concatenating them, which actually gave better results (for classification), but I am not sure this needs to be supported.

tadej-redstone:
Adds `layer_index` param to both transformer encoders. Closes #1284

### DIFF
--- a/encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
+++ b/encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
@@ -64,6 +64,12 @@ encoders_parameters = [
         "pooling_strategy": 'max',
         "pretrained_model_name_or_path": 'xlnet-base-cased',
         "model_save_path": 'xlnet-base-cased-max',
+    },
+    {
+        "pooling_strategy": 'mean',
+        "pretrained_model_name_or_path": 'bert-base-uncased',
+        "model_save_path": 'bert-base-uncased-mean',
+        "layer_index": -2,
     }
 ]
 
@@ -111,3 +117,10 @@ def test_parameter_override(encoder):
     assert encoder.pretrained_model_name_or_path == encoder_preset['pretrained_model_name_or_path']
     assert encoder.pooling_strategy == encoder_preset['pooling_strategy']
     assert encoder.model_save_path == encoder_preset['model_save_path']
+
+@pytest.mark.parametrize('encoder', encoders_parameters[0:1], indirect=['encoder'])
+def test_wrong_layer_index(encoder):
+    encoder.layer_index = 100
+    test_data = np.array(['it is a good day!', 'the dog sits on the floor.'])
+    with pytest.raises(ValueError, match=f'Invalid value {encoder.layer_index}'):
+        encoded_data = encoder.encode(test_data)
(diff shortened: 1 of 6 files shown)


JoanFM:
Hey @tadej-redstone, thanks for the contribution.

But please raise the version manifest of the executors for the tests to really be executed.

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -111,3 +117,10 @@ def test_parameter_override(encoder):
     assert encoder.pretrained_model_name_or_path == encoder_preset['pretrained_model_name_or_path']
     assert encoder.pooling_strategy == encoder_preset['pooling_strategy']
     assert encoder.model_save_path == encoder_preset['model_save_path']
+
+@pytest.mark.parametrize('encoder', encoders_parameters[0:1], indirect=['encoder'])
```
JoanFM, line 121: can we test also a different layer than -1 and see that it indeed works? And also test what happens if we use -100 as an edgecase?

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
tadej-redstone: Please check the the tests. I have added a tests for layer `-2`, as well as a test for layer `100` (which should raise an error)

tadej-redstone: Sorry, or did you mean you want to also test -100, as well as 100?

JoanFM: ah yes sorry did not realize the test. It would be cool to test -100 although it may be overkilling. What is really important is to increase the versions in manifest.yml

tadej-redstone: I have done that now :)

tadej-redstone:
@JoanFM The tests are now failing because the CI is using v4.0 of transformers. I have fixed that in #1285, so can we merge that in first and then re-run the CI here?

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
JoanFM: where is the test?

JoanFM: ah I see you changed the range, can there be a test with large negative value? It is good also as documentation to know edge cases behavior

tadej-redstone: Ok, added test for -100 as well

JoanFM (changes requested):
Now you need to bump the manifest version again in this PR

JoanFM:
There are several conflicts right now you need to care for

JoanFM (approved):

PR #1289, opened 2020-12-13
fix: train_filepath can be none and loaded from binary

### LINKED ISSUE #1288: Loading binary files with FaissIndexer._load_training_data
The BaseNumpyIndexer.get_create_handler method stores vectors in gzip files. When the compression-level is zero, the vectors are stored as plain binary files. This case causes a problem when loading the training-data for the FaissIndexer.

Thus, I suggest that the FaissIndexer._load_training_data method should also support binary files. 
```
# Read from binary file:
with open(self.train_filepath, 'rb') as f:
    train_data = f.read()
```
This would enable us to train the FaissIndexer on the indexed vectors.

In addition, I'd like to mention that it was really hard for me to figure out how to use a plain FlatIndex (no training) for Faiss. Basically you just have to use "Flat" as you index_key and set is_trained to true in the meta. Maybe you want to add this to the documentations or examples.

All the best
Jan

JoanFM:
**Changes introduced**
Fixes #1288 .

- Does not force compression level
- Allows having no loading data if not needed.
- Allow training data from uncompressed data if required

### DIFF
--- a/indexers/vector/FaissIndexer/__init__.py
+++ b/indexers/vector/FaissIndexer/__init__.py
@@ -1,7 +1,7 @@
 __copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-from typing import Tuple
+from typing import Tuple, Optional
 
 import numpy as np
 from jina.executors.devices import FaissDevice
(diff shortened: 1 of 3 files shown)


nan-wang (approved):
LGTM👍

PR #1300, opened 2020-12-16
ci: specify comparison level

cristianmtr:
Add comparison level for when new Jina releases should be tested.


JoanFM:
@cristianmtr any context on this PR? Can you add the details on why this is needed?

cristianmtr:
> @cristianmtr any context on this PR? Can you add the details on why this is needed?

Yes, thanks for noticing. Atm the Hub Updater is only testing if the minor versions match, but we want to also test on patch releases. See https://github.com/jina-ai/action-hub-updater/pull/1/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R18

JoanFM (approved):

Issue #1296, opened 2020-12-15
Refactor Transformer tests


bhavsarpratik:
Refactor TransformerTorch and TransformerTF tests with a matrix of parameters by using `pytest parametrize`. The matrix should contain a few popular models worth testing with different `pooling` and `layer_index`

tadej-redstone:
I can take care of this - I'm working on some other refactoring of transformer encoders anyways.

JoanFM:
> I can take care of this - I'm working on some other refactoring of transformer encoders anyways.

That would be very cool. What refactoring are u working on?

tadej-redstone:
> > I can take care of this - I'm working on some other refactoring of transformer encoders anyways.
> 
> That would be very cool. What refactoring are u working on?

What I described in #1291, some minor changes in documentation to synchronise with transformers package, and implementing reduce operations in native pytorch/tf - as this can lead to large speedups on GPU - same as in #399

tadej-redstone:
And I am also considering adding AMP, for pytorch at least - see n #1283

bhavsarpratik:
closed by #1344

JoanFM:
This is not done in `TFTransformerEncoder` right?

bhavsarpratik:
Ohh yes!

bhavsarpratik:
@tadej-redstone Would you be interested in doing similar refactoring for `TransformerTFEncoder`? It shouldn't take much time.

tadej-redstone:
Yes, I already have a branch ready doing this, I'm waiting first till #1527
is fixed

Issue #1291, opened 2020-12-14
Remove truncation_strategy parameter from transformer encoders


tadej-redstone:
This parameter is completely unneeded for the transformer encoders. Here's why:
- different truncation options (`only_first`, `only_second`, `longest_first`) make no sense for us. They are only different if you are passing in two sentences separated by the `SEP` token - this is used only in the cases like multiple choice or question answering. We *never* use the `SEP` token.
- The no truncation option (`False`) can lead to two possible behaviors: either all the inputs are below max permissible model size, and this is the same as `True`, or an input is above max model size, and the whole thing crashes when we try to encode these tokenized sentences. So there's no real reason why anyone would want to set this to `False`. 

So we should simply set `truncation_strategy` to `True`. We can deprecate this parameter first

PR #1344, opened 2020-12-20
perf: improve performance of TransformerTorchEncoder and refactor code

tadej-redstone:
I've improved performance of TransformerTorchEncoder and refactored the module + tests code. To show that the changes I did really improved performance, I created [this notebook](https://colab.research.google.com/drive/1PQJmvgnwotbcbZleLZJXOCvML9sIxwAu?usp=sharing) where I demonstrate the effect of each change. I'll list the changes below:

### Performance changes

- The `padding` parameter for tokenization is no longer `max_length`, but `longest` (this only affects padding, not truncation).  The first section of the notebook demonstrates that this can lead to large speedups in encoding, up to 10x! And this has a negligible affect on results (max difference is on the order of 1e-3 percent of std)
- The reductions (`mean`, `max`, `cls`...) are no longer done in numpy, but in pytorch. On CPU the performance improvement is from ~25% for short sequences, and up to 30x for long sequences. On GPU, the performance increase goes up to 65x. Some of the increase is due to the new code simply being more efficient than before (and possibly pytorch using multiprocessing on CPU?), while for GPU I show that a lot of this increase comes from the fact that if the reductions are done on the GPU, less data needs to be transferred between GPU and CPU.

### Refactoring

- As described in #1291, I deprecate the `truncation_strategy` parameter, as the different strategies are all the same for us.
- I also deprecate the `"auto"` pooling strategy - as I see it, it is exactly equal to `"cls"`, and as it is no longer even the default, I don't see why it is needed.
- Description of `pretrained_model_name_or_path` parameter is updated to macth the description that is now in the huggingface library.
- I introduce the `base_tokenizer_model` parameter for the model for the tokenizer (defaults to `pretrained_model_name_or_path`, so no breaking changes). The reason for this is that if you fine tune and save your own transformer model, this model will still use the vocabulary of the model that it is based on - unless you altered the vocabulary and want to save it explicitly. In my use cases, I never used/saved custom vocabularies, so I think that the option to use standard tokenizers is needed.
- `AutoModel` is used instead of `AutoModelForPretraining` - they are the same, except that `AutoModelForPretraining` has an additional pretraining head (for maskedLM/causualLM/...) which we weren't using anyway.
- I refactor tests in accordance with #1296.


encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -8,170 +8,156 @@
...
 class TransformerTorchEncoder(TorchDevice, BaseEncoder):
     """
-    Internally, TransformerTorchEncoder wraps the tensorflow-version of transformers from huggingface.
+    Internally, TransformerTorchEncoder wraps the pytorch version of transformers from huggingface.
     """
 
     def __init__(
         self,
         pretrained_model_name_or_path: str = 'sentence-transformers/distilbert-base-nli-stsb-mean-tokens',
+        base_model: str = None,
```
JoanFM, line 23: if it is None, then the correct hint would be Optional]

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -8,170 +8,156 @@
...
+            ..deprecated::
+                This parameter has no effect - ``truncation=True`` is always applied on tokenizing by default.
 
         ..warning::
             `model_save_path` should be relative to executor's workspace
         """
 
         super().__init__(*args, **kwargs)
         self.pretrained_model_name_or_path = pretrained_model_name_or_path
+        self.base_model = base_model if base_model else pretrained_model_name_or_path
```
JoanFM, line 52: Not sure, but is possible that `base_model or pretrained_model_name_or_path`  would also work?

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -8,170 +8,156 @@
...
     def tokenizer(self):
         from transformers import AutoTokenizer
-        tokenizer = AutoTokenizer.from_pretrained(self.pretrained_model_name_or_path)
+        tokenizer = AutoTokenizer.from_pretrained(self.base_model)
         return tokenizer
 
     @batching
     @as_ndarray
-    def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
+    def encode(self, data: "np.ndarray", *args, **kwargs) -> "np.ndarray":
```
JoanFM, line 105: please keep single quote in the code as per our style

encoders/nlp/TransformerTorchEncoder/__init__.py
JoanFM: would `base_tokenizer_model` be a more informative name?

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -8,170 +8,156 @@
...
 
     def __init__(
         self,
         pretrained_model_name_or_path: str = 'sentence-transformers/distilbert-base-nli-stsb-mean-tokens',
+        base_model: str = None,
         pooling_strategy: str = 'mean',
         layer_index: int = -1,
         max_length: Optional[int] = None,
-        truncation_strategy: str = 'longest_first',
+        truncation_strategy: bool = True,
```
JoanFM, line 27: `strategy` seems a weird name for a boolean option right?

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -1,140 +1,144 @@
...
-@pytest.fixture(scope='function')
-def test_metas(tmpdir, random_workspace_name):
-    os.environ[random_workspace_name] = str(tmpdir)
+@pytest.fixture(scope="function")
+def test_metas(tmp_path):
     metas = get_default_metas()
-    metas['workspace'] = os.environ[random_workspace_name]
-    if 'JINA_TEST_GPU' in os.environ:
-        metas['on_gpu'] = True
+    metas["workspace"] = str(tmp_path)
```
JoanFM, line 17: same here, keep single quote

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
tadej-redstone: Will do. I was actually trying to find any reference about the code style, but didn't find anything. Is it available somewhere?

encoders/nlp/TransformerTorchEncoder/__init__.py
tadej-redstone: True, I can probably just leave it at `'longest_first'`, as it's deprecated anyways

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -1,140 +1,144 @@
...
+]
+encoder_params_ind_1 = filter(lambda x: x["layer_index"] == -1, encoder_params)
+
+
+def _assert_params_equal(params_dict: dict, encoder: TransformerTorchEncoder):
+    for key, val in params_dict.items():
+        assert val == getattr(encoder, key)
+
+
+@pytest.mark.parametrize("encoder", encoder_params, indirect=True)
```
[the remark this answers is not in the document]
JoanFM: This will make the test much more readable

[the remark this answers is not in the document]
tadej-redstone: The problem is, if I do this, then I have to put a statement like
``` python
encoder = Torchencoder(
    param_1=param_1,
    param_2=param_2,
    ...
)
```

in basically every test, as there is no way to use pytest fixtures indirectly for multiple params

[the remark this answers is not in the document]
JoanFM: not sure if is possible but I still prefer this because it makes more explicit than the need to understand how the encoder was built. Plus if a test fails it has in the test name the combination of parameters.

JoanFM:
Bad commit message, please refer to:
- [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
- [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -1,140 +1,165 @@
...
-    assert encoder_loaded.max_length == encoder.max_length
     np.testing.assert_array_equal(encoded_data_control, encoded_data_test)
 
 
-@pytest.mark.parametrize('encoder', encoders_parameters, indirect=['encoder'])
-def test_save_and_load_config(encoder):
+@pytest.mark.parametrize('model_name', _models)
+@pytest.mark.parametrize('pooling_strategy', ['cls', 'mean', 'max'])
+@pytest.mark.parametrize('layer_index', [-1, -2, 0])
+def test_save_and_load_config(test_metas, model_name, pooling_strategy, layer_index):
```
JoanFM, line 125: no need this huge matrix of parameters for these save and load tests

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -1,140 +1,165 @@
...
+
+    np.testing.assert_allclose(encoded_data[0], encoded_data[1], atol=1e-5, rtol=1e-4)
 
 
-@pytest.mark.parametrize('encoder', encoders_parameters, indirect=['encoder'])
-def test_save_and_load(encoder):
+@pytest.mark.parametrize('model_name', _models)
+@pytest.mark.parametrize('pooling_strategy', ['cls', 'mean', 'max'])
+@pytest.mark.parametrize('layer_index', [-1, -2, 0])
+def test_save_and_load(test_metas, model_name, pooling_strategy, layer_index):
```
JoanFM, line 96: no need of this big matrix of parameters

tadej-redstone:
> Bad commit message, please refer to:
> 
> * [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
> * [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)

Which commit message?

JoanFM:
> > Bad commit message, please refer to:
> > 
> > * [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
> > * [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)
> 
> Which commit message? Edit: nvm, i see

See error in CI:

```
Run wagoid/commitlint-github-action@v1
/usr/bin/docker run --name wagoidcommitlintgithubaction180_6618e7 --label cc4956 --workdir /github/workspace --rm -e GITHUB_TOKEN -e INPUT_CONFIGFILE -e INPUT_FIRSTPARENT -e INPUT_FAILONWARNINGS -e INPUT_HELPURL -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_ACTION_REPOSITORY -e GITHUB_ACTION_REF -e GITHUB_PATH -e GITHUB_ENV -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/_temp/_runner_file_commands":"/github/file_commands" -v "/home/runner/work/jina-hub/jina-hub":"/github/workspace" wagoid/commitlint-github-action:1.8.0
Error: You have commit messages with errors

⧗   input: tests: use parametrize to build tests matrix
✖   type must be one of [build, chore, ci, docs, feat, fix, improvement, perf, refactor, revert, style, test] [type-enum]
```

tadejsv:
@JoanFM the CI is now passing

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -8,120 +8,96 @@
...
     """
 
     def __init__(
         self,
         pretrained_model_name_or_path: str = 'sentence-transformers/distilbert-base-nli-stsb-mean-tokens',
+        base_tokenizer_model: Optional[str] = None,
         pooling_strategy: str = 'mean',
         layer_index: int = -1,
         max_length: Optional[int] = None,
         truncation_strategy: str = 'longest_first',
```
bhavsarpratik, line 27: Currently, this is an unused variable.

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -1,140 +1,161 @@
...
-        "model_save_path": 'distilbert-base-uncased-mean',
-        "layer_index": -2,
-    },
-    {
-        "pooling_strategy": 'mean',
-        "pretrained_model_name_or_path": 'distilbert-base-cased',
-        "model_save_path": 'distilbert-base-cased-mean',
-        "max_length": 100
-    },
+
```
bhavsarpratik, line 18: We need to add a test for checking the quality of emb as there has been a lot of changes. Take emb of the `bert-base-uncased` using the old code and save it as npy. Use it to assert the new code produces same emb. (keep old code defaults such that they match with current)

bhavsarpratik (changes requested):

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
tadej-redstone: Sure, will do

tadej-redstone: @bhavsarpratik done, tests pass :)

JoanFM (approved):

bhavsarpratik:
Thanks a lot @tadej-redstone for this contribution 👍

bhavsarpratik (approved):

PR #1347, opened 2020-12-22
refactor: move transformer model loading to post_init

tadejsv:
According to [BaseExecutor documentation](https://github.com/jina-ai/jina/blob/master/jina/executors/__init__.py#L219-L231), model loading should be done in `post_init`:

>  Initialize class attributes/members that can/should not be (de)serialized in standard way.
>  Examples:
>             **- deep learning models**
>             - index files
>             - numpy arrays

So I moved the loading of model/tokenizer for Transformer Encoders to `post_init`.

This also makes sense since jina logs the time taken in post_init and warns the user that this might take some time - while no such logging/warning is given for model loading outside of post_init, as it is done now

### DIFF
--- a/encoders/nlp/TransformerTFEncoder/__init__.py
+++ b/encoders/nlp/TransformerTFEncoder/__init__.py
@@ -98,12 +98,14 @@ class TransformerTFEncoder(TFDevice, BaseEncoder):
         """
         return self.get_file_from_workspace(self.model_save_path)
 
-    @cached_property
-    def model(self):
-        from transformers import TFAutoModelForPreTraining
-        model = TFAutoModelForPreTraining.from_pretrained(self.pretrained_model_name_or_path)
+    def post_init(self):
+        from transformers import TFAutoModel, AutoTokenizer
+
+        self.tokenizer = AutoTokenizer.from_pretrained(self.base_tokenizer_model)
+        self.model = TFAutoModel.from_pretrained(
+            self.pretrained_model_name_or_path, output_hidden_states=True
+        )
         self.to_device()
-        return model
 
     @cached_property
     def no_gradients(self):
@@ -114,12 +116,6 @@ class TransformerTFEncoder(TFDevice, BaseEncoder):
     def tensor_func(self):
         import tensorflow as tf
         return tf.constant
-
-    @cached_property
-    def tokenizer(self):
-        from transformers import AutoTokenizer
-        tokenizer = AutoTokenizer.from_pretrained(self.pretrained_model_name_or_path)
-        return tokenizer
 
     @batching
     @as_ndarray
@@ -145,9 +141,7 @@ class TransformerTFEncoder(TFDevice, BaseEncoder):
         token_ids_batch = self.array2tensor(ids_info['input_ids'])
         mask_ids_batch = self.array2tensor(ids_info['attention_mask'])
         with self.no_gradients():
-            outputs = self.model(token_ids_batch,
-                                 attention_mask=mask_ids_batch,
-                                 output_hidden_states=True)
+            outputs = self.model(token_ids_batch, attention_mask=mask_ids_batch)
 
             hidden_states = outputs[-1]
             n_layers = len(hidden_states)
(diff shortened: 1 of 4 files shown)


encoders/nlp/TransformerTFEncoder/__init__.py
```diff
@@ -145,9 +141,7 @@ def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
         token_ids_batch = self.array2tensor(ids_info['input_ids'])
         mask_ids_batch = self.array2tensor(ids_info['attention_mask'])
         with self.no_gradients():
-            outputs = self.model(token_ids_batch,
-                                 attention_mask=mask_ids_batch,
-                                 output_hidden_states=True)
```
JoanFM, line 150: is `output_hidden_states` not needed anymore?

JoanFM (commented):
I like the change because it is more clear, but just for info,

the main point of having a `post_init` method is to avoid loading expensive objects in `__init__` because `jina` has a mechanism to avoid serializing the `objects` initialized in `post_init`. This behavior is equivalent to the `cached_property` objects which are also skipped in serialization.

encoders/nlp/TransformerTFEncoder/__init__.py
tadejsv: You can see that `output_hidden_states` is now set in `post_init`

JoanFM: very good point ;)

tadejsv:
@JoanFM there's one mysterious error in CI with config not saving properly for TF (not related to anything changed in this PR) - any idea what to do about it?

tadejsv:
I see the exact same error appears [here](https://github.com/jina-ai/jina-hub/pull/1349), and also locally these tests fail for me. Should I file an issue about it with more details?

JoanFM:
> I see the exact same error appears [here](https://github.com/jina-ai/jina-hub/pull/1349), and also locally these tests fail for me. Should I file an issue about it with more details?

that would be good

JoanFM:
> @JoanFM there's one mysterious error in CI with config not saving properly for TF (not related to anything changed in this PR) - any idea what to do about it?

It is indeed very weird failure

tadejsv:
@JoanFM do you think we could move forward with merging this PR (if other tests pass) despite that strage bug? It seems not be related with anything changed in this PR (as it also happens on the automatic update bot PRs...)

JoanFM:
> @JoanFM do you think we could move forward with merging this PR (if other tests pass) despite that strange bug? It seems not be related with anything changed in this PR (as it also happens on the automatic update bot PRs...)

Hey @tadejsv ,

Sorry but it can't be merged, I recommend you split it in 2 PRs one for each executor. Merging this would mean there is no alignment between the images in `jinahub` and the code in the repository. I will try to dig deeper in the root cause

tadejsv:
@JoanFM, okay, I can do that. Btw, related to images - it doesn't seem like any of the hub images have been updated [in the last few months](https://hub.docker.com/u/jinaai)

JoanFM:
> @JoanFM, okay, I can do that. Btw, related to images - it doesn't seem like any of the hub images have been updated [in the last few months](https://hub.docker.com/u/jinaai)

No, they are built constantly https://hub.docker.com/u/jinahub

tadejsv:
> > @JoanFM, okay, I can do that. Btw, related to images - it doesn't seem like any of the hub images have been updated [in the last few months](https://hub.docker.com/u/jinaai)
> 
> No, they are built constantly https://hub.docker.com/u/jinahub

Oh okay, then docs need to be updated

JoanFM:
> > > @JoanFM, okay, I can do that. Btw, related to images - it doesn't seem like any of the hub images have been updated [in the last few months](https://hub.docker.com/u/jinaai)
> > 
> > 
> > No, they are built constantly https://hub.docker.com/u/jinahub
> 
> Oh okay, then docs need to be updated

Sure, can you point where the link was found?

tadejsv:
@JoanFM yes, for example here: https://docs.jina.ai/chapters/hub/publish-your-pod-image.html and here: https://docs.jina.ai/chapters/hub/use-your-pod.html

But there are some other pages in the documentation with the wrong link too, just search for `jinaai/hub.` and you'll fine them.
The first page (about publishing your pod) actually needs a more thorough update, as the structure of `manifest.yml` doesn't correspond to the actual one, and neither does the structure of the files inside `jina-hub`.

JoanFM:
Hey @tadejsv ,

 TransformerTFEncoder can be updated now!

PR #1758, opened 2021-01-03
feat: include cmlm encoder

carlosb1:


encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -30,11 +34,28 @@ def __init__(
...
     def post_init(self):
         self.to_device()
+        if UNIVERSAL_SENTENCE_ENCODER_CMLM == self.model_url:
+            self.model = self._load_model_cmlm
+
+        else:
+            self.model = self._load_model_universal_sentence
+
+    @as_ndarray
+    def _load_model_cmlm(self, data: 'np.ndarray') -> 'np.ndarray':
```
JoanFM, line 44: can't this be unified? Just having an encodrr and an optional preprocessor and have the user input them?

JoanFM (changes requested):
Please bump the version in the manifest.yml for the tests to be run

florian-hoenicke:
Hi @carlosb1,
thanks for your contribution. 
Please make sure your commit messages have the right format.
Here are further information:
https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md
Your commit messages can be changed using `git rebase -i HEAD~<N>`

encoders/nlp/UniversalSentenceEncoder/__init__.py
carlosb1: do you talk about  avoiding  these multiple functions and reuse the .load function? If it is this, I can do it and if you want but I am not sure if it would open for an extension (close-open principle), I am only encapsulating your different use cases when you are initializing a model. Anyway, it doesn't exit more use cases (we need preprocess and we don't need it), it seems easier solution.

JoanFM: but the point is to have this as a wrapper for this KerasLayer and preprocessing (if possible) and then have the user choose which encoder from TF Hub can be used and which preprocessor (if any). 

I am not sure about the details but I undersatnd this pattern should work for different encoders of the same family.

encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -16,8 +20,8 @@ class UniversalSentenceEncoder(BaseTFEncoder):
 
     def __init__(
             self,
-            model_url: str = 'https://tfhub.dev/google/universal-sentence-encoder/4',
-            *args,
+            model_url: str = UNIVERSAL_SENTENCE_ENCODER,
```
JoanFM, line 23: The point would be:

to have 

```
def __init__(self, 
                    model_url: str = 'https://tfhub.dev/google/universal-sentence-encoder/4',
                    preprocessor_url: Optional[str]= None, 
                    *args, 
                   **kwargs):
....
self.preprocessor_url = preprocessor_url
self.model_url = model_url
```

and in post_init:

```
def post_init(self):
       import tensorflow_hub as hub
       preprocessor = None
       if preprocessor_url:
            preprocessor = hub.KerasLayer(preprocessor_url)
        encoder = hub.KerasLayer(model_url)
        return encoder(preprocessor(data))['default'] if preprocessor else encoder(data)['default']
```

Is that possible to work for any `universal-sentence-encoder` family?

encoders/nlp/UniversalSentenceEncoder/__init__.py
carlosb1: I will change it... It is not related with a pattern, it is more a design principle, but the point is that you have clear your requirement, you want to specify your preprocessor outside an "optional" parameter, then the natural way is to do it as you want.

encoders/nlp/UniversalSentenceEncoder/__init__.py
carlosb1: I think the return  line is not for this method.

carlosb1:
```
Error: You have commit messages with errors

⧗   input: fix: set fixed versions requirements except for jina in some encoders (#1784)
✖   header must not be longer than 72 characters, current length is 77 [header-max-length]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

⧗   input:  fix: remove unused code (#1783)
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```
Is it possible that this error come from some previous PR?

JoanFM:
> ```
> Error: You have commit messages with errors
> 
> ⧗   input: fix: set fixed versions requirements except for jina in some encoders (#1784)
> ✖   header must not be longer than 72 characters, current length is 77 [header-max-length]
> 
> ✖   found 1 problems, 0 warnings
> ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
> 
> ⧗   input:  fix: remove unused code (#1783)
> ✖   subject may not be empty [subject-empty]
> ✖   type may not be empty [type-empty]
> 
> ✖   found 2 problems, 0 warnings
> ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
> ```
> 
> Is it possible that this error come from some previous PR?

Hey @carlosb1 , very likely,

Because this branch seems that has done some weird rebases, I see 175 changes in this PR?

Maybe is better to start in anothe branch and reopen this PR

PR #1780, opened 2021-01-04
fix: set fixed versions requirements except for jina

JoanFM:


cristianmtr (approved):

crafters/image/TorchObjectDetectionSegmenter/requirements.txt
```diff
@@ -1,2 +1,2 @@
 torchvision>=0.3.0
```
cristianmtr, line 1: Shouldn't `torchvision` also be a fixed requirement?

cristianmtr (approved):
Shouldn't `torchvision` also be a fixed requirement?

crafters/image/TorchObjectDetectionSegmenter/requirements.txt
JoanFM: oh yes, missed that

PR #1782, opened 2021-01-04
fix: set fixed versions requirements except for jina encoders part 2

JoanFM:


encoders/nlp/LaserEncoder/aux.py
```diff
@@ -0,0 +1,22 @@
+import pkg_resources
+
+def test(requirements_path):
```
cristianmtr, line 3: I'm not sure I understand how this would be used. Is it part of CI, or is it up to the dev to run it?

cristianmtr: Why are we setting that version? Can't we just do it in the requirements?

JoanFM: oh this is a file that should not have been uploaded will remove

encoders/nlp/FarmTextEncoder/requirements.txt
```diff
@@ -1 +1 @@
-farm
+farm==0.5.0
```
florian-hoenicke, line 1: can optionally be set to `0.6.0`. The tests pass.

florian-hoenicke (dismissed):

encoders/nlp/UniversalSentenceEncoder/requirements.txt
```diff
@@ -1,4 +1,4 @@
-protobuf>=3.9.2
-grpcio>=1.24.3
-tensorflow
-tensorflow_hub
\ No newline at end of file
+protobuf==3.14.0
+grpcio==1.34.0
```
maximilianwerk, line 2: Why are they sometimes removed and sometimes bumped? Shouldn't they come from core directly?

JoanFM: when they are removed is because some tests were failing and dig deeper to fix it.

JoanFM: trying to remove noew

PR #1786, opened 2021-01-04
fix: fix versions for indexers and rankers

JoanFM:


indexers/vector/FaissIndexer/requirements.txt
```diff
@@ -1 +1 @@
-faiss-cpu
+faiss-cpu==1.6.3
```
florian-hoenicke, line 1: does `faiss-cpu==1.6.5` work as well?

JoanFM: I do not know, I checked the current version and fixed the version that is currently installed, I can't go further and innovate fromt hat

florian-hoenicke: Pycharm was suggesting me `1.6.5` and now I tried it out. All tests pass.

JoanFM: Done!

florian-hoenicke (approved):

PR #1804, opened 2021-01-05
fix: fix vsetext encoder versions

JoanFM:


encoders/nlp/VSETextEncoder/requirements.txt
```diff
@@ -1,4 +1,4 @@
 jina
```
cristianmtr, line 1: This is because it's using `FROM pytorch/pytorch:latest`, right?

cristianmtr (approved):

encoders/nlp/VSETextEncoder/requirements.txt
JoanFM: yes

PR #1781, opened 2021-01-04
fix: set fixed versions requirements except for jina in encoders

JoanFM:

PR #1903, opened 2021-01-07
fix: fix grpc conflicts

nan-wang:
fix: fix grpc conflicts


JoanFM:
This is being handled in the PR #1781

nan-wang:
> This is being handled in the PR #1781

yes, the thing is that this blocks the CI. :(

JoanFM:
yes, I changed that, the point is if we merge this, u will cancel the effect of my PR since the manifest will be not updated, let me close for now

PR #1904, opened 2021-01-07
fix: avoid warning when importing

JoanFM:


cristianmtr (approved):
Strange: the git diff is completely confused by the indentation change. Enabling `Hide whitespace changes` in GH diff viewer made it much clearer to understand

PR #1905, opened 2021-01-07
fix: fix faiss indexer train everytime

JoanFM:


nan-wang (approved):
good for now. But we need a better way to fix this. `is_trained` is always loading the default value from the `executors.metas.default.yml`

JoanFM:
> good for now. But we need a better way to fix this. `is_trained` is always loading the default value from the `executors.metas.default.yml`

Yes, but the point is that `is_trained` is supposed to be used for the `train` interfaces and now they do not play any role

PR #2070, opened 2021-01-10
feat: add cmlm encoder

carlosb1:
It adds universal cmlm encoder

### DIFF
--- a/encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
+++ b/encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
@@ -75,3 +75,23 @@ def test_save_and_load_config(mocker):
     encoder_loaded = BaseExecutor.load_config(encoder.config_abspath)
     assert encoder_loaded.model_url == encoder.model_url
     rm_files([encoder.config_abspath])
+
+
+@mock.patch('tensorflow_hub.load', return_value=MockModule())
+def test_get_universal_sentence_encoder(mocker):
+    MODEL_ENCODER = 'https://tfhub.dev/google/universal-sentence-encoder/4'
+    metas = get_metas()
+    encoder = UniversalSentenceEncoder(metas=metas)
+    assert encoder.model_url == MODEL_ENCODER
+
+
+@mock.patch('tensorflow_hub.load', return_value=MockModule())
+def test_get_universal_sentence_encoder_cmlm(mocker):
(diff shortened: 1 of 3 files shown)


carlosb1:
@JoanFM , I am not sure, but it doesn't seem related with my rebase. I did another one, (it seems updated with `git log`) and it keeps throwing the same error.

JoanFM (changes requested):
Hey @carlosb1 ,

it looks really good, I think it is the idea to have a potential preprocessor

I do however would like to see a test where a preprocessor is applied.

Also, please bump the manifest version so that the CI and the tests are fired

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
```diff
@@ -75,3 +75,23 @@ def test_save_and_load_config(mocker):
...
+@mock.patch('tensorflow_hub.load', return_value=MockModule())
+def test_get_universal_sentence_encoder(mocker):
+    MODEL_ENCODER = 'https://tfhub.dev/google/universal-sentence-encoder/4'
+    metas = get_metas()
+    encoder = UniversalSentenceEncoder(metas=metas)
+    assert encoder.model_url == MODEL_ENCODER
+
+
+@mock.patch('tensorflow_hub.load', return_value=MockModule())
+def test_get_universal_sentence_encoder_cmlm(mocker):
```
carlosb1, line 89: @JoanFM  Here we have a test to validate the preprocess parameter. do we need another one?

JoanFM:
Hey @carlosb1 , it seems a problem of a last commit that I squashed, I guess you trying to rebase is firing the problem again? Can you maybe create a fresh directly from master and copying these changes?

Sorry for the inconvinience

carlosb1 (commented):
modified

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
JoanFM: I mean that it actually properly encodes some input, here we are only testing initialization and post_init.

carlosb1:
> Hey @carlosb1 ,
> 
> it looks really good, I think it is the idea to have a potential preprocessor
> 
> I do however would like to see a test where a preprocessor is applied.
> 
> Also, please bump the manifest version so that the CI and the tests are fired

The error should be there.

PR #2158, opened 2021-01-11
refactor: improve torch transformer encoder dockerfile + add gpu

tadej-redstone:
This creates brand new docker file for the torch transformer encoder. 

I am now basing it on conda, as that is the recommended way to install pytorch. I have also figured out why the container is so large (currently almost 6GB!) - the test artifacts were not properly removed. I've fixed that here.

Additionally, I've added a GPU dockerfile, which is almost identical to the CPU one. Future PR needs to make sure that the GPU dockerfile is built and uploaded to the hub


encoders/nlp/TransformerTorchEncoder/requirements.txt
```diff
@@ -1,2 +1 @@
-transformers==4.1.1
-torch==1.7.1
+jina
```
JoanFM, line 1: why can't the dependencies be kept here?

encoders/nlp/TransformerTorchEncoder/gpu.Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM continuumio/miniconda3 AS conda
```
JoanFM, line 1: would this Dockerfile also work for non-gpu environments?

tadej-redstone: yes it would - but there would be a lot of "bloat". The GPU image will end up being ~3GB due to CUDA/torch, while i think the CPU should be about only 300MB

encoders/nlp/TransformerTorchEncoder/requirements.txt
tadej-redstone: This way both CUDA and torch can be installed together very easily. This way of installing is also the recommended way listed on pytorch website.

tadej-redstone:
@JoanFM the same issue with params not being saved in config as in [1527](https://github.com/jina-ai/jina/issues/1527) appeared in CI, can you help with this issue?

encoders/nlp/TransformerTorchEncoder/Dockerfile
```diff
@@ -1,13 +1,30 @@
...
 # setup the workspace
 COPY . /workspace
 WORKDIR /workspace
 
-# install the third-party requirements
-RUN pip install -r requirements.txt
-
 # for testing the image
-RUN pip install pytest && pytest && rm -rf '~/.cache/huggingface/transformers/'
+RUN pip install pytest && pytest && rm -rf '/root/.cache/huggingface/transformers/' && rm -rf /tmp/{*,.*}
```
JoanFM, line 28: can u remove also env.yml and requirements.txt from the root folder (that are copied at the begining? to avoid duplication?

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/Dockerfile
tadejsv: They are not part of the image - only the /opt folder is copied from the conda image, everything else is discarded

JoanFM:
Hey @tadejsv  may u rebase with master since a PR has been merged taking that version number?

tadejsv:
@JoanFM can you restart the CI here? should work with 0.9.19

JoanFM (approved):

PR #2159, opened 2021-01-11
fix: clean tf transformer encoder artifacts

tadej-redstone:
properly cleans transformer test artifacts, should reduce the image size by a few GB :)


JoanFM (approved):

tadej-redstone:
Yup, size went from 3.6 to 1.54 GB :)

PR #2179, opened 2021-01-12
refactor: shadowing built ins

florian-hoenicke:
refactor: shadowing built ins

### DIFF
--- a/indexers/vector/MilvusIndexer/milvusdbhandler.py
+++ b/indexers/vector/MilvusIndexer/milvusdbhandler.py
@@ -98,11 +98,11 @@ class MilvusDBHandler:
             db.insert(reduce(operator.concat, keys.tolist()), vectors)
 
     def build_index(self, index_type: str, index_params: dict):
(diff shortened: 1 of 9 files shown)


JoanFM (changes requested):
please remember to bump image versions, and do not change in Milvus because it is broken and will not be fixed

indexers/vector/MilvusIndexer/milvusdbhandler.py
```diff
@@ -98,11 +98,11 @@ def insert(self, keys: 'np.ndarray', vectors: 'np.ndarray'):
             db.insert(reduce(operator.concat, keys.tolist()), vectors)
 
     def build_index(self, index_type: str, index_params: dict):
-        t = self.get_index_type(index_type)
+        type = self.get_index_type(index_type)
```
JoanFM: remove changes with Milvus because it cannot be updated and at least there will be a mapping between last published image and actual code

JoanFM (approved):

PR #2202, opened 2021-01-13
feat:  new cmlm encoder

carlosb1:
It adds new cmlm sentence encoder.


JoanFM (changes requested):
can you please bump the version to 0.0.8?

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
```diff
@@ -75,3 +77,23 @@ def test_save_and_load_config(mocker):
...
+@mock.patch('tensorflow_hub.KerasLayer', return_value=MockModule())
+def test_get_universal_sentence_encoder(mocker):
+    MODEL_ENCODER = 'https://tfhub.dev/google/universal-sentence-encoder/4'
+    metas = get_metas()
+    encoder = UniversalSentenceEncoder(metas=metas)
+    assert encoder.model_url == MODEL_ENCODER
+
+
+@mock.patch('tensorflow_hub.KerasLayer', return_value=MockModule())
+def test_get_universal_sentence_encoder_cmlm(mocker):
```
JoanFM, line 91: Please add some encoding here asserting some property of the resulting embedding to make sure it is tested with preprocessors

encoders/nlp/UniversalSentenceEncoder/manifest.yml
```diff
@@ -7,7 +7,7 @@ author: Jina AI Dev-Team (dev-team@jina.ai)
 url: https://jina.ai
 vendor: Jina AI Limited
 documentation: https://github.com/jina-ai/jina-hub
-version: 0.0.7
+version: 0.0.1
```
JoanFM, line 10: please bump the version to the next version 0.0.8 and not 0.0.1

JoanFM (changes requested):

encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -17,21 +19,29 @@ class UniversalSentenceEncoder(BaseTFEncoder):
...
         :param kwargs:
         """
         super().__init__(*args, **kwargs)
         self.model_url = model_url
+        self.preprocessor_url = preprocessor_url
 
     def post_init(self):
         self.to_device()
+        import tensorflow as tf
+        import tensorflow_text as text
```
nan-wang, line 39: Unused import?
```suggestion
```

nan-wang (dismissed):

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
```diff
@@ -34,6 +35,7 @@ def __call__(self, data, *args, **kwargs):
 target_output_dim = 512
 test_data = np.array(['it is a good day!', 'the dog sits on the floor.'])
 
+
 def _test_encoding_results():
```
JoanFM, line 39: we could parametrize this test to wotk with and without `preprocessing`

JoanFM (changes requested):

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
carlosb1: I see another problem...The workflow cmlm implies to select a result dictionary 'default'
 `english_embeds = encoder(preprocessor(english_sentences))["default"]`
but for the old code it is used 
`model.load()`
As far I know, It is necessary to split in different two use cases and it can not be reused how I pretend in the current PR. For me, it seems easier to create two different private functions for each case.

encoders/nlp/UniversalSentenceEncoder/__init__.py
carlosb1: It fails if it is not added...
` FileNotFoundError: Op type not registered 'CaseFoldUTF8' in binary running on `

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
carlosb1: About the test... I didn t get it.. do you want to add tests for the validation of ouput preprocessor, is it correct  ?

encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -17,21 +19,29 @@ class UniversalSentenceEncoder(BaseTFEncoder):
...
         import tensorflow_hub as hub
-        self.model = hub.load(self.model_url)
+        self.preprocessor = None
```
carlosb1, line 41: I see another problem...The workflow cmlm implies to select a result dictionary 'default'
 `english_embeds = encoder(preprocessor(english_sentences))["default"]`
but for the old code it is used 
`model.load()`
As far I know, It is necessary to split in different two use cases and it can not be reused how I pretend in the current PR. For me, it seems easier to create two different private functions for each case.

encoders/nlp/UniversalSentenceEncoder/tests/test_universalsentenceencoder.py
JoanFM: I want to add test that the encoding also works when there is preprocessing.

carlosb1: ok and  e2e test to verify that it is all working well. not problem.

encoders/nlp/UniversalSentenceEncoder/__init__.py
JoanFM: and how would u detect each use case?

carlosb1: In the end, you have two different pipelines that  depend on your input model (your type of encoder). My usual methodology to do this, it keeps all the time simple the code (KISS principle) and understand what it is defining these two use cases. As far I know, it is defined by a string path model.  Also, for having  the control of this path model, I would define them as a constant... you can  manage which  input variables are correct in the function...

JoanFM: would it then make sense to add a totally brand new Executor? to habdle the different cases? I am afraid this will fall into having a lot of if ... else ...

carlosb1: In the end, we are doing baby steps for refactoring... Yes, in the first iteration you will detect this technical debt with this one if / else....   but also , you have a general view of the pipeline... If you have a function with a if / else that its only responsability is tocall  two functions (no more), now, you understand that you class has two responsibilities / logic concepts (break the single responsability pattern), you need a `polymorphism` [pattern](https://stackoverflow.com/questions/47402390/polymorphism-vs-if-and-logic)... Which one? you can start to think with of your design patterns can fit better in your use case... 
Sorry if it is a big comment for a PR... Also... not forget all this redesign / refactor principles  are a fight between technical debt and clean code, you can keep it "dirty" waiting for the best moment to change it.

JoanFM: Well the polymorphism exists from the BaseExecutor class all the way down to the hub executors. So here the only point is where these 2 logics will become so complex that the need for having another Executor is more clear.

You can try to do the `if ... else` logic and see how it looks.

As per the technical debt, it is not such a problem ih the `hub` as this is a `library` of executors specially thought to have a clear and small interface (`init`, `post_init`, `encode`) where one can fit its own logic. So as long as the usage and testing of this specific class is clear there should not be a problem

carlosb1: Polymorphism could be applicated with a [composition](https://en.wikipedia.org/wiki/Composition_over_inheritance) structure also. It didn t review yet all your code, but it is a possibility.

JoanFM: yes but we sacrifice polymorphism and code reuse to have no dependency between hub executors as much as possible so that they have no dependencies and can be seen as independent small projects

carlosb1: I think you don t need to sacrifice this with a composition structure... but I admit I would need to check deeper the code to provide you a proposal...

JoanFM: if it does not involve affecting other executors in the hub it would be okey

carlosb1:
* I added two private class to apply a delegate pattern for splitting responsabilties remove ifs (how it was discussed)
* I used the model name to define which encoder can be used (How I was discussed)
* This delegation can permit for future PRs, if it is necessary the extraction of these private classes
* For a next iteration, in can be done a polymorph pattern if we use multiple inheritance
* I added two types of tests, one with and without mocks to verify the creation of the encoders

encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -26,21 +81,25 @@ def __init__(
...
+        if self.model_url == MODEL_ENCODER_CMLM:
+            self.sentence_encoder = CMLMEncoder()
+        else:
+            self.sentence_encoder = GeneralEncoder(self.model_url)
+
     def post_init(self):
         self.to_device()
-        import tensorflow_hub as hub
-        self.model = hub.load(self.model_url)
+        self.sentence_encoder.post_init()
```
JoanFM, line 94: You should directly instantiatr the private classes here, otherwise they will be serialized. Also no need to have post_init pattern in the private classes.

encoders/nlp/UniversalSentenceEncoder/__init__.py
```diff
@@ -1,12 +1,67 @@
...
+
+import tensorflow_hub as hub
 import numpy as np
 
 from jina.executors.decorators import batching, as_ndarray
 from jina.executors.encoders.frameworks import BaseTFEncoder
 
 
+UNIVERSAL_SENTENCE_ENCODER = 'https://tfhub.dev/google/universal-sentence-encoder/4'
+MODEL_ENCODER_CMLM = "https://tfhub.dev/google/universal-sentence-encoder-cmlm/en-base/1"
```
JoanFM, line 14: This is coming from my lack of knowledge, but there is only one possible combination of Model URL and preprocessor URL that deserve handling?

JoanFM (changes requested):

JoanFM:
> * I added two private class to apply a delegate pattern for splitting responsabilties remove ifs (how it was discussed)
> * I used the model name to define which encoder can be used (How I was discussed)
> * This delegation can permit for future PRs, if it is necessary the extraction of these private classes
> * For a next iteration, in can be done a polymorph pattern if we use multiple inheritance
> * I added two types of tests, one with and without mocks to verify the creation of the encoders

Looks good!

encoders/nlp/UniversalSentenceEncoder/__init__.py
JoanFM: would it make sense to have a boolean parameter (cmlm) or (masked model) to decide wether a preprocessing must be applied? or there exists only one single model with this API?

carlosb1: Sorry, I know you asked this before I admit, I am not sure.. I tried to do it open for an extension if  in some near future we need to include this.

carlosb1: https://tfhub.dev/google/collections/bert/1 <- In fact, it seems possible, we can include the feature now

carlosb1: I can verify the paper, if it  is necessary.

JoanFM: We can go with this for now I think. I think we just need to move the instantiatiom of private classes to post init, and no need to keep 'post_init` in the private classes

JoanFM:
recheckcla

carlosb1:
> **Jina CLA check**
> 
>  Thank you for your pull request. It looks like this is your first contribution to an open source project maintained by Jina AI Limited. Before we can look at your pull request, we kindly ask that you sign our [Contributor License Agreement](https://github.com/jina-ai/jina/blob/master/CLA.md). You can sign it by commenting in format below.
> 
> _**I have read the CLA Document and I hereby sign the CLA**_

This web doen't work for me.

JoanFM:
> > **Jina CLA check**
> > Thank you for your pull request. It looks like this is your first contribution to an open source project maintained by Jina AI Limited. Before we can look at your pull request, we kindly ask that you sign our [Contributor License Agreement](https://github.com/jina-ai/jina/blob/master/CLA.md). You can sign it by commenting in format below.
> > _**I have read the CLA Document and I hereby sign the CLA**_
> 
> This web doen't work for me.

We just need you to write **I have read the CLA Document and I hereby sign the CLA** in a comment in this PR in case you agree

JoanFM:
Terms to be found here https://github.com/jina-ai/legal/blob/master/CLA.md

carlosb1:

JoanFM (approved):
Thank you very much for the great contribution!

PR #2347, opened 2021-01-14
feat: add training max number of points to consider

JoanFM:


nan-wang (approved):
The multi-if part needs further refactoring. I'll merge it atm

PR #2565, opened 2021-01-18
fix: update multimodal encoder

bwanglzu:
Update positional modality for the multimodal encoder, and update text model url.


JoanFM (changes requested):
please bump version so that it is rebuilt

JoanFM (approved):

PR #2765, opened 2021-01-24
fix: imageresizer docker branch

carlosb1:
The Dockefile for the imageresizer crafter uses an old branch, removed.


carlosb1:
some idea? I only removed one tag and merged from master. Also, I don't see any error message.

JoanFM (commented):
on master the devel branch is already removed, so no need to change it there? were u testing with an old version of the code?

carlosb1:
yes, my fault, I close this.

PR #2933, opened 2021-01-27
reformat: clean transformers

tadej-redstone:
This removes model/tokenizer saving from transformer encoders.

They are not needed as the models are simply used pretrained, no training (modification) is done.


JoanFM (changes requested):
Ok now I understand why it was done, can  you please bump the version?

JoanFM (approved):

PR #3047, opened 2021-01-31
feat: add gpu dockerfile

nan-wang:
- [ ] fix the cuda 11.0 dependency issue. https://www.tensorflow.org/install/gpu#install_cuda_with_apt
- [ ] test the usage of gpu pod


nan-wang:
This issue is stale and closed because it has been open for more than 20 days with no activity.

PR #3049, opened 2021-01-31
docs: sptagindexer crud

florian-hoenicke:


indexers/vector/SptagIndexer/__init__.py
```diff
@@ -94,6 +98,10 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
             return _index
 
     def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+        """Find the top-k vectors with smallest ``metric`` and return their ids in ascending order.
```
JoanFM, line 101: have u checked if this applies to every metric or if there is an exception? for instance with FaissIndexer the inner product returns similarity, (maybe we should transform to distance ourselves)

JoanFM (commented):
are u sure you are aligned with latest master? https://github.com/jina-ai/jina-hub/pull/3501/files

JoanFM (commented):
ah ok I see now

indexers/vector/SptagIndexer/Dockerfile
```diff
@@ -12,10 +12,8 @@ RUN wget https://cmake.org/files/v3.15/cmake-3.15.5.tar.gz && \
     make -j2 && \
     make install
 
-RUN git clone https://github.com/microsoft/SPTAG.git
+RUN git clone https://github.com/microsoft/SPTAG.git/tree/a91ed49a041a24bd145874d3b3c815bde0430e65
```
JoanFM, line 15: can we merge these 2 RUN layers?

florian-hoenicke: I thought so but have to reverse the commit

indexers/vector/SptagIndexer/__init__.py
florian-hoenicke: Good point.
I checked the documentation:
`Vectors returned for a query vector are the vectors that have smallest L2 distance or cosine distances with the query vector.`
Only cosine and l2 are supported.
Also, I added an assertion to the test to make sure the order is ascending.

theUnkownName (approved):

maximilianwerk (approved):
LGTM

PR #3055, opened 2021-01-31
Docs milvusindexer crud

florian-hoenicke:


JoanFM (commented):
MilvusIndexer will not work yet. Also it does not inherit from BaseNumpyIndexer and would need to have crud operations added.

However the Milvus API is not stable and we will adapt to it later

florian-hoenicke:
Thanks, I closed the pr.

PR #3501, opened 2021-02-05
fix: sptag indexer compile issue

rutujasurve94:
- Use compatible commit `a91ed49a041a24bd145874d3b3c815bde0430e65` from `SPTAG` repo
- Fixes compilation issues for `SptagIndexer`


indexers/vector/SptagIndexer/Dockerfile
```diff
@@ -13,8 +13,10 @@ RUN wget https://cmake.org/files/v3.15/cmake-3.15.5.tar.gz && \
     make install
 
 RUN git clone https://github.com/microsoft/SPTAG.git
-RUN cd SPTAG && \
-    mkdir build && \
+RUN cd SPTAG
```
JoanFM, line 16: please do this in a single `RUN` so that not so many layers are created

JoanFM (changes requested):

indexers/vector/SptagIndexer/tests/test_sptagindexer.py
```diff
@@ -75,7 +75,7 @@ def test_sptag_indexer_known(metas):
     with BaseIndexer.load(save_abspath) as indexer:
         assert isinstance(indexer, SptagIndexer)
         idx, dist = indexer.query(queries, top_k=2)
-        np.testing.assert_equal(idx, np.array([[4, 5], [5, 4], [6, 5], [7, 6]]).astype(str))
+        np.testing.assert_equal(idx, np.array([[4.0, 5.0], [5.0, 4.0], [6.0, 5.0], [7.0, 6.0]]).astype(str))
```
JoanFM, line 78: It is weird because they should be str

JoanFM (changes requested):

indexers/vector/SptagIndexer/tests/test_sptagindexer.py
```diff
@@ -75,18 +75,18 @@ def test_sptag_indexer_known(metas):
     with BaseIndexer.load(save_abspath) as indexer:
         assert isinstance(indexer, SptagIndexer)
         idx, dist = indexer.query(queries, top_k=2)
-        np.testing.assert_equal(idx, np.array([[4, 5], [5, 4], [6, 5], [7, 6]]))
+        np.testing.assert_equal(idx, np.array([[4., 5.], [5., 4.], [6., 5.], [7., 6.]]))
```
JoanFM, line 78: this seems weird to me, it should be equal to the `np.array` as type str

JoanFM (changes requested):

indexers/vector/SptagIndexer/tests/test_sptagindexer.py
rutujasurve94: the `_int2ext_id` method is implemented in `BaseNumpyIndexer` and not `BaseIndexer`

indexers/vector/SptagIndexer/tests/test_sptagindexer.py
```diff
@@ -75,18 +75,18 @@ def test_sptag_indexer_known(metas):
     with BaseIndexer.load(save_abspath) as indexer:
         assert isinstance(indexer, SptagIndexer)
         idx, dist = indexer.query(queries, top_k=2)
-        np.testing.assert_equal(idx, np.array([[4, 5], [5, 4], [6, 5], [7, 6]]))
+        np.testing.assert_equal(idx.astype(str), np.array([[4.0, 5.0], [5.0, 4.0], [6.0, 5.0], [7.0, 6.0]]).astype(str))
```
cristianmtr, line 78: But why do you need to convert to `str`? These should be strings from the return

rutujasurve94: If I don't convert `lhs` (`idx`) to `str` explicitly, I get:
```
          HubIO@21[I]:E           Mismatched elements: 1 / 1 (100%)
          HubIO@21[I]:E            x: array([[4, 5],
          HubIO@21[I]:E                  [5, 4],
          HubIO@21[I]:E                  [6, 5],
          HubIO@21[I]:E                  [7, 6]])
          HubIO@21[I]:E            y: array([['4.0', '5.0'],
          HubIO@21[I]:E                  ['5.0', '4.0'],
          HubIO@21[I]:E                  ['6.0', '5.0'],
          HubIO@21[I]:E                  ['7.0', '6.0']], dtype='<U32')
```
As `unicode` comparison with `numeric` datatypes won't returned equal

cristianmtr: Yes, I understand. 

The problem is that SptagIndexer behaves differently. See Annoy here : https://github.com/jina-ai/jina-hub/blob/998a881b80841b47ba170c2f6cd1496a7aeb1b6d/indexers/vector/AnnoyIndexer/tests/test_annoyindexer.py#L100

You can see that in this case we are only converting to `str` the right-hand element in the comparison. That's what I would expect here as well. 

@JoanFM Any ideas why this is happening?

JoanFM: exactly, u should not need to cast `idx`

JoanFM: also u should not have it as `float`

JoanFM (changes requested):

rutujasurve94:
@JoanFM @cristianmtr if I don't cast `idx` as string and `np.array` with float, I get `u 32` != `u 21` error as by default `idx` returns `u 32` and our non-float `np.array` is `u 21`

```
          HubIO@46010[I]:E           AssertionError: 
          HubIO@46010[I]:E           Arrays are not equal
          HubIO@46010[I]:E           
          HubIO@46010[I]:E           Mismatched elements: 8 / 8 (100%)
          HubIO@46010[I]:E            x: array([['4.0', '5.0'],
          HubIO@46010[I]:E                  ['5.0', '4.0'],
          HubIO@46010[I]:E                  ['6.0', '5.0'],
          HubIO@46010[I]:E                  ['7.0', '6.0']], dtype='<U32')
          HubIO@46010[I]:E            y: array([['4', '5'],
          HubIO@46010[I]:E                  ['5', '4'],
          HubIO@46010[I]:E                  ['6', '5'],
          HubIO@46010[I]:E                  ['7', '6']], dtype='<U21')
          HubIO@46010[I]:
```
This is very weird

I suspect this has got to do with `SpTag` itself

cristianmtr (approved):

JoanFM (approved):

PR #3537, opened 2021-02-08
docs: redis interfaces

florian-hoenicke:


indexers/keyvalue/RedisDBIndexer/__init__.py
```diff
@@ -1,10 +1,13 @@
...
+from typing import Optional, Iterable
 
 from jina.executors.indexers.keyvalue import BinaryPbIndexer
 
+if False:
+    from redis import Redis
+
 
 class RedisDBIndexer(BinaryPbIndexer):
     """
```
theUnkownName, line 13: Shouldn't we add a description for all classes?

indexers/keyvalue/RedisDBIndexer/__init__.py
```diff
@@ -56,7 +57,7 @@ def add(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
             for k in redis_docs:
                 redis_handler.set(k['_id'], k['values'])
 
-    def update(self, keys: Iterator[int], values: Iterator[bytes], *args, **kwargs):
+    def update(self, keys: Iterable[str], values: Iterable[bytes], *args, **kwargs) -> None:
         """Update JSON-friendly serialized documents on the index.
 
         :param keys: document ids to update
```
theUnkownName, line 63: I think it needs a description of all params along with their types

theUnkownName (changes requested):

indexers/keyvalue/RedisDBIndexer/__init__.py
florian-hoenicke: not needed as discussed

JoanFM (approved):

PR #3548, opened 2021-02-09
docs: docstrings for segmenters/NLP

theUnkownName:


segmenters/nlp/Sentencizer/__init__.py
```diff
@@ -8,8 +8,19 @@
...
     :class:`Sentencizer` split the text on the doc-level into sentences on the chunk-level with a rule-base strategy.
-        The text is split by the punctuation characters listed in ``punct_chars``.
-        The sentences that are shorter than the ``min_sent_len`` or longer than the ``max_sent_len`` after stripping will be discarded.
+    The text is split by the punctuation characters listed in ``punct_chars``.
+    The sentences that are shorter than the ``min_sent_len`` or longer than the ``max_sent_len`` after stripping will be discarded.
+
+    :param min_sent_len: the minimal number of characters,
+        (including white spaces) of the sentence, by default 1.
+    :param max_sent_len: the maximal number of characters,
+        (including white spaces) of the sentence, by default 1e5.
```
JoanFM, line 17: 1e5 is no the valid default right?

segmenters/nlp/Sentencizer/__init__.py
```diff
@@ -8,8 +8,19 @@
...
+    :param punct_chars: the punctuation characters to split on.
```
JoanFM, line 18: I think for `punct_chars` may be good to show an example, (like ',.') is `,`  OR `.`

JoanFM (changes requested):

segmenters/nlp/Sentencizer/__init__.py
theUnkownName: But we have so many, should we add just a few on the docstring?

JoanFM: what I mean is that the argument I do not find it so intuitive to say that by giving a string with different punctuaction chars ',.:' means that any of the characters in the string will be used to split a document

theUnkownName: What you think of this: '

`the punctuation characters to split on, for example, but not limited to '.', ',', '!'`

segmenters/nlp/Sentencizer/__init__.py
```diff
@@ -14,8 +14,9 @@ class Sentencizer(BaseSegmenter):
     :param min_sent_len: the minimal number of characters,
         (including white spaces) of the sentence, by default 1.
     :param max_sent_len: the maximal number of characters,
-        (including white spaces) of the sentence, by default 1e5.
-    :param punct_chars: the punctuation characters to split on.
+        (including white spaces) of the sentence, by default 512.
+    :param punct_chars: the punctuation characters to split on,
```
JoanFM, line 18: Actually, the problem is that the `type` hint is wrong, it should be `Optional[Tuple[str]]`

JoanFM (changes requested):

segmenters/nlp/Sentencizer/__init__.py
theUnkownName: What you mean? for `max_sent_len`? this is an `int` right?

JoanFM: the punct_chars I mean

theUnkownName: Ah, but isn't the type `List[str]`?

JoanFM: well it says it is `str`

theUnkownName: Yeah that's wrong, but I was wondering for the type `Optional[Tuple[str]]` you mentioned

JoanFM: Optional[List[str]] is good also. Optional indicates that a value of None is properly handled qnd since None is the default parameter let's add it here

JoanFM (commented):
LGTM, u need to make sure to bump the versions for the tests to pass

JoanFM (approved):

PR #3647, opened 2021-02-10
docs: docstring for rankers

theUnkownName:


JoanFM (approved):

rankers/BM25Ranker/__init__.py
```diff
@@ -9,42 +9,52 @@
...
     """
     required_keys = {'length', }
 
     def __init__(self, k=1.2, b=0.75, threshold=0.1, *args, **kwargs):
-        """
-
-        :param k: the parameter ``k`` for the BM25 algorithm.
-        :param b: the parameter ``b`` for the BM25 algorithm.
-        """
+        """Set constructor."""
```
maximilianwerk, line 31: Remove this docstring completely. According to our updated guidelines, `__init__` function docstring should completely go into the class definition.

maximilianwerk (changes requested):
Small formal change

rankers/BM25Ranker/__init__.py
theUnkownName: Yes all the params of the contructor have to be on the class, but we still need a docstring specifying this for the constructor right? As in their docs: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

maximilianwerk: According to the sphinx documentations (https://www.sphinx-doc.org/ar/stable/usage/extensions/example_google.html#example-google) you either do class and `__init__` documentation or everything in the class docstring. You don't mix it (e.g. putting parameters of `__init__` into class and still write something for the constructor itself).

theUnkownName: @maximilianwerk but that's for the Google style (using Sphinx, but not "sphinx style") right? I can remove it but then I get a warning from the `pydocstyle`, are we ok with that?

maximilianwerk: OK, you convinced me - let's keep it. I start to understand why google style might be the way to go at some point in time.

theUnkownName: Could you approve the changes to merge? :)

maximilianwerk (approved):

PR #3675, opened 2021-02-12
fix: scann indexer

florian-hoenicke:


indexers/vector/ScannIndexer/tests/test_scannindexer.py
```diff
@@ -26,7 +26,7 @@ def metas(tmpdir):
 
 def test_scannindexer(metas):
     with ScannIndexer(index_filename='scann.test.gz', metas=metas) as indexer:
-        indexer.add(vec_idx, vec)
+        indexer.add([str(e) for e in vec_idx], vec)
```
JoanFM, line 29: change the type of vecidx

JoanFM (changes requested):

JoanFM (commented):
vec_idx = np.random.randint(0, high=100, size=[10])

This should be casted to `astype(str)`

indexers/vector/ScannIndexer/tests/test_scannindexer.py
florian-hoenicke: true, I made it more readable. Thanks

JoanFM (approved):

Issue #3068, opened 2021-02-01
Create Zero-Short transformer based classifiers


tadej-redstone:
I think a great addition to the hub would be a zero-shot classifier. You can read more about them [on this link](https://joeddav.github.io/blog/2020/05/29/ZSL.html), but in general they work like this:

1. Each of the possible categories is represented as a string (e.g. "Financial Document", "Legal Document", ...), and is transformed into an embedding
2. The object to categorize is also represented by a string ("This contract concerns the relationship between..."), which is then transformed into embeddings.
3. A distance metric is computed between the embeddings of categories and the item to categorize. From here it's simple to obtain a probability distribution on the categories and then return the assigned labels.

Their advantage is that they require zero training and can be adapted to a wide variety of classification tasks. They (some of the more sophisticated variants) also achieved quite great performance. For this reason I think they would fit in well in the Jina flow.

There are actually different interesting variants of this (as explained in the linked blog post), such as using a cloze task for classification, or, in case of single word categories, adjusting them to the embedding space through word2vec embeddings.

I think this is a great task for a first time contributor - the classifier would build upon the transformer encoders, but otherwise after some time I could take a shot at it as well.

JoanFM:
Hey @tadej-redstone ,

This could be interesting as part of our `Classifier` BaseExecutor.  As per the interface described it would be a `classifier` assigning labels to each document. Currently we do not have any of these in our Hub but it can be a good addition.

As per the dependency on `TransformerEncoder`, one design decision of our Hub Executors is that they should be independent of each other even when code duplication may be involved. So even when sharing a lot of code with `TrasformerEncoders` they should be written from scratch.

nan-wang:
I love the idea

bsridatta:
Sounds interesting, I would like to contribute to this

deepampatel:
I was using `ZeroShotClassifier` in my side project. 
Currently the way `BasePredictDriver` is implemented it requires the classification to be done using the embeddings generated by an encoder which is ahead in the flow.

gateway -> craft -> encode -> classify ....

The way i did it was i implemented a `BasePredictDriver` that used `extract_docs` to get text, then predict labels using zero shot classifier.

  ```
class CustomPredictDriver(BasePredictDriver):

      def _apply_all(
              self,
              docs: Iterable['jina_pb2.Document'],
              context_doc: 'jina_pb2.Document',
              field: str,
              *args,
              **kwargs,
      ) -> None:
[...]

PR #3753, opened 2021-02-15
feat: add ZeroShotTFClassifier

julianpetrich:
Adds wrapper for TF transformers to do zero-shot classification


JoanFM (commented):
I like the changes, but as discussed in #3068, the current Drivers do not accept `text` as input. So first we should change the drivers, otherwise this `Classifier` will not be useful

JoanFM (changes requested):
I see a lot of tests focusing on the encoding part, but not on the `prediction` part which is the public method of the class and its main interface.

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if self.pooling_strategy not in ["cls", "mean", "max", "min"]:
+            self.logger.error(
+                f"pooling strategy not found: {self.pooling_strategy}."
+                ' The allowed pooling strategies are'
+                '"cls", "mean", "max", "min".'
+            )
+            raise NotImplementedError
+
+    def predict(self,
```
JoanFM, line 70: This interface does not respect the `interface` of our `BaseClassifier`. It should only classify fron `data`, there should not be any dependency with `target_labels`

JoanFM (changes requested):

classifiers/nlp/ZeroShotTFClassifier/__init__.py
julianpetrich: Changed it to be part of the keywords. For zero shot classification we do need the argument in some form as the classifier will classify based on distance.

julianpetrich:
> I see a lot of tests focusing on the encoding part, but not on the `prediction` part which is the public method of the class and its main interface.

Good point, I have created a few more tests for the prediction part.

julianpetrich:
> I like the changes, but as discussed in #3068, the current Drivers do not accept `text` as input. So first we should change the drivers, otherwise this `Classifier` will not be useful

I will look into the driver later this week as it is already very late here :)

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -69,7 +69,8 @@ def __init__(
 
     def predict(self,
                 data: "np.ndarray",
-                target_labels: "np.ndarray") -> "np.ndarray":
+                *args,
```
JoanFM, line 72: if `target_labels` are needed they should be defined apart from `*args`, `**kwargs`. 

But anyway, right now if this is needed we may need to rethink the interface, since there is no Driver in `jina` that would be able to work with this.

JoanFM (changes requested):

classifiers/nlp/ZeroShotTFClassifier/tests/test_zeroshottfclassifier.py
```diff
@@ -0,0 +1,307 @@
...
+        classifier._encode(test_data)
+
+
+def test_wrong_pooling_strategy():
+    with pytest.raises(NotImplementedError):
+        ZeroShotTFClassifier(pooling_strategy="wrong")
+
+
+@pytest.mark.parametrize(
+    "params",
```
JoanFM, line 214: could you please use `single quotes`?

classifiers/nlp/ZeroShotTFClassifier/__init__.py
JoanFM: I think what should happen is that `target_labels` should be an attribute of the `instance` and not part of the `predict` interface. So it will be a `Classifier` for a closed set of `classes`

JoanFM (changes requested):

classifiers/nlp/ZeroShotTFClassifier/manifest.yml
```diff
@@ -0,0 +1,13 @@
+manifest_version: 1
+name: ZeroShotTFClassifier
+kind: classifier
+type: pod
+description: :class:`ZeroShotTFClassifier` wraps tensorflow transformers from huggingface,
+  and performs zero shot learning classification for nlp data.
+author: Jina AI Dev-Team (dev-team@jina.ai)
```
nan-wang, line 7: hi, @julianpetrich We'd suggest to put your github account name here. In the future, we plan to provide a GUI for managing all your hub images at one place.

```suggestion
author: julianpetrich
```

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+                  actual: "np.ndarray",
+                  desired: "np.ndarray") -> "np.ndarray":
+
+        actual = _expand_vector(actual)
+        desired = _expand_vector(desired)
+
+        return _cosine(_ext_A(_norm(actual)), _ext_B(_norm(desired)))
+
+    @batching
+    @as_ndarray
```
nan-wang, line 133:
```suggestion

```

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+        if len(self.labels) < 2:
+            raise ValueError("The number of target labels must be at least 2.")
+
+        if len(self.labels) != \
+                len(set(self.labels)):
+            raise ValueError(
+                "There are duplicate value in the target_label argument."
+            )
+
+    def predict(self,
```
nan-wang, line 83:
```suggestion
    @batching
    @as_ndarray
    def predict(self,
```

nan-wang (approved):

classifiers/nlp/ZeroShotTFClassifier/README.md
```diff
@@ -0,0 +1,4 @@
+# ZeroShotTFClassifier
```
JoanFM, line 1: Hey @julianpetrich ,

We would need some better and more insightful Readme for this specific executor, with examples on how to use and a little more explanation of the parameters and the inner logic.

You can see how these changes are being applied to the rest of executors

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+        pretrained_model_name_or_path: str = "distilbert-base-uncased",
+        base_tokenizer_model: Optional[str] = None,
+        pooling_strategy: str = "mean",
+        layer_index: int = -1,
+        max_length: Optional[int] = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        :param labels: the potential labels for the classification
```
JoanFM, line 29: Our docstring guidelines suggest to put the `__init__` params at the class level and not in the constructor.

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    :class: `ZeroShotTFClassifier` wraps tensorflow transformers from
+        huggingface, and performs zero shot learning classification
+        for nlp data.
+    """
+
+    def __init__(
+        self,
+        labels: List[str],
+        pretrained_model_name_or_path: str = "distilbert-base-uncased",
```
JoanFM, line 20: Could u try to use `single-quote` wherever possible (also in tests and comments and raise messages)

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _encode(self, data: "np.ndarray", *args, **kwargs) -> "np.ndarray":
+
+        import tensorflow as tf
+
+        if not self.tokenizer.pad_token:
+            self.tokenizer.add_special_tokens({"pad_token": "[PAD]"})
+            self.model.resize_token_embeddings(len(self.tokenizer.vocab))
+
+        input_tokens = self.tokenizer(
```
JoanFM, line 142: are we sure we are handling properly the input if the model works in GPU?

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+         :param data: the input textual data to be classified, a 1 d
+            array of string type in size `B`
+         :type data: np.ndarray
+         :return: zero/one one-hot predicted label of each sample
+            in size `(B, L)`
+         :rtype: np.ndarray
+        """
+
+        data_encoded = self._encode(data)
+        labels_encoded = self._encode(self.labels)
```
JoanFM, line 106: since the `labels` are constant, no need to encode them every time, I think in `post_init` u can keep an array of `encodings` for all the labels as an optimization

JoanFM (changes requested):

classifiers/nlp/ZeroShotTFClassifier/__init__.py
nan-wang: In the requirement, it is specified to use `tensorflow-cpu` and therefore we don't need to consider GPU usages.

JoanFM: but as many other executors, but considering it can help us to prepare for its usage.

Plus some users may prefer to install their dependencies and use it without dockee

classifiers/nlp/ZeroShotTFClassifier/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+**`ZeroShotTFClassifier`** wraps the TensorFlow version of transformers from 
+huggingface, and performs zero-shot learning classification for NLP data. 
+
+It takes an ndarray of sequences and return a ndarray of labels.
+
+* Input shape: `BatchSize`
```
JoanFM, line 8: you may mean length and not shape?

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+
+        if self.pooling_strategy not in ['cls', 'mean', 'max', 'min']:
+            self.logger.error(
+                f'pooling strategy not found: {self.pooling_strategy}.'
+                ' The allowed pooling strategies are'
+                ' `cls`, `mean`, `max`, `min`.'
+            )
+            raise NotImplementedError
+
+        if len(self.labels) < 2:
```
JoanFM, line 72: doesnt it work to have a yes/No prediction with a single label?

classifiers/nlp/ZeroShotTFClassifier/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+            raise ValueError('The number of target labels must be at least 2.')
+
+        if len(self.labels) != \
+                len(set(self.labels)):
+            raise ValueError(
+                'There are duplicate value in the target_label argument.'
+            )
+
+    @batching
```
JoanFM, line 81: I am not sure, but I think is better to swap the order of these decorators, so that the as_nparray applies to the full output

JoanFM (changes requested):

classifiers/nlp/ZeroShotTFClassifier/__init__.py
julianpetrich: No, it doesn't. It checks the distance from one sequence to each potential class, and then labels the sequence depending on the min distance. With one label it would label each sequence with that one label.

JoanFM:
recheckcla

classifiers/nlp/ZeroShotTFClassifier/README.md
JoanFM:
```suggestion
* Input length: `BatchSize`
```

JoanFM (dismissed):
just a minimal thing

julianpetrich:

nan-wang (approved):
LGTM👍

PR #3872, opened 2021-02-17
feat: integrate image metadata in document

kujta1:
Each image contains textual and numerical information that can be used for searching and can help to have a better ranking. These information are of different aspects like, Geolocation, photogramteric, hardware specification as well as other characteristic information are usually called and saved as metadata. 
For example, if you just want to search for photos that have been shot in Winter, you can get this information from the metadata. If you only want to search for photos shot with a Samsung S20 phone , you can get this information from metadata, or GPS of the photo. All modern digital cameras save this information, some of them even the depth map of the image. 
I think similar relevant  information  can be extracted from audio/video and pdf files.

### DIFF
--- a/crafters/image/ImageReader/__init__.py
+++ b/crafters/image/ImageReader/__init__.py
@@ -39,4 +39,23 @@ class ImageReader(BaseCrafter):
         img = np.array(raw_img).astype('float32')
         if self.channel_axis != -1:
             img = np.moveaxis(img, -1, self.channel_axis)
-        return dict(weight=1., blob=img)
+        # Add image metadata
+        from PIL.ExifTags import TAGS
+        from PIL.ExifTags import GPSTAGS 
+        # Extract image metadata
+        # https://github.com/python-pillow/Pillow/blob/master/src/PIL/ExifTags.py
+        metadata = {}
+        exif_data = raw_img.getexif()
+        for tag_id in exif_data:
+            tag = TAGS.get(tag_id)
+            if tag is None:
+                tag = GPSTAGS.get(tag_id, tag_id)
+            data = exif_data.get(tag_id)
+            # transform bytes to readable data
+            if isinstance(data, bytes):
+                data = data.decode()
+            # if tag does not exists in Pillow
+            if tag is None:
+                continue
+            metadata[tag] = data
+        return dict(weight=1., blob=img, metadata=metadata)
--- a/crafters/image/ImageReader/tests/test_imagereader.py
+++ b/crafters/image/ImageReader/tests/test_imagereader.py
@@ -34,3 +34,12 @@ def test_io_buffer():
     test_doc = crafter.craft(buffer=image_buffer.getvalue(), uri=None)
     assert test_doc['blob'].shape == (img_size, img_size, 3)
     np.testing.assert_almost_equal(test_doc['blob'], np.array(img).astype('float32'))
+
+
+def test_io_buffer_metadata():
+    crafter = ImageReader()
+    import requests
+    img_url='https://raw.githubusercontent.com/x4nth055/pythoncode-tutorials/master/ethical-hacking/image-metadata-extractor/image.jpg'
+    test_doc = crafter.craft(buffer=None, uri=requests.get(img_url, stream=True).raw)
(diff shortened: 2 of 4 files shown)


crafters/image/ImageReader/__init__.py
```diff
@@ -39,4 +39,23 @@ def craft(self, buffer: bytes, uri: str, *args, **kwargs) -> Dict:
         img = np.array(raw_img).astype('float32')
         if self.channel_axis != -1:
             img = np.moveaxis(img, -1, self.channel_axis)
-        return dict(weight=1., blob=img)
+        # Add image metadata
+        from PIL.ExifTags import TAGS
```
JoanFM, line 43: while I really like the feature, I think it would be good to have this activated with an instance `argument`.

crafters/image/ImageReader/tests/test_imagereader.py
```diff
@@ -34,3 +34,12 @@ def test_io_buffer():
     test_doc = crafter.craft(buffer=image_buffer.getvalue(), uri=None)
     assert test_doc['blob'].shape == (img_size, img_size, 3)
     np.testing.assert_almost_equal(test_doc['blob'], np.array(img).astype('float32'))
+
+
+def test_io_buffer_metadata():
+    crafter = ImageReader()
+    import requests
+    img_url='https://raw.githubusercontent.com/x4nth055/pythoncode-tutorials/master/ethical-hacking/image-metadata-extractor/image.jpg'
```
JoanFM, line 42: if this `image` is OpenSource, I would suggest to download it and add it inside the `tests` folder. I think is better not to depend on texternal Urls that may eventually change and may harm the test

JoanFM (changes requested):
Hey @kujta1 ,

I think is quite a cool feature, could you take a look at the comments and let me know what you think?

crafters/image/ImageReader/tests/test_imagereader.py
kujta1: hi @JoanFM  I have taken the image from this repo: https://github.com/x4nth055/pythoncode-tutorials which has an MIT licence, but I am not entirely sure about the [image.jpg](https://github.com/x4nth055/pythoncode-tutorials/tree/master/ethical-hacking/image-metadata-extractor), because of that I was afraid to put it on the repo.

crafters/image/ImageReader/__init__.py
```diff
@@ -39,4 +39,23 @@ def craft(self, buffer: bytes, uri: str, *args, **kwargs) -> Dict:
...
+        from PIL.ExifTags import GPSTAGS 
+        # Extract image metadata
+        # https://github.com/python-pillow/Pillow/blob/master/src/PIL/ExifTags.py
+        metadata = {}
+        exif_data = raw_img.getexif()
+        for tag_id in exif_data:
+            tag = TAGS.get(tag_id)
+            if tag is None:
+                tag = GPSTAGS.get(tag_id, tag_id)
```
JoanFM, line 52: what is this logic here? why passing tag_id twice?

crafters/image/ImageReader/tests/test_imagereader.py
JoanFM: I think we may find some proper test image data that has metadata and can be valid for us.

JoanFM: Also it would be good to add mocking tests to ensure that every option of the metadata info is tested, for instance the check if data is of type bytes

JoanFM (changes requested):

crafters/image/ImageReader/__init__.py
kujta1: From [Pillow Code](https://github.com/python-pillow/Pillow/blob/master/src/PIL/ExifTags.py) we have two dictionaries (TAGS and GPSTAGS), which transform each tag ID into a human-readable string. The TAGS dictionary has tags that are general characteristics (features). GPSTAGS are specifically related to GPS locations.

rutujasurve94:
@kujta1 any updates on this?

nan-wang:
I will close this PR for the time being.

PR #3879, opened 2021-02-18
fix: keras image encoder respects model name

maximilianwerk:
The `model_name` parameter never worked for the `ImageKerasEncoder`.


cristianmtr (approved):

JoanFM (changes requested):
may we add a test with different model names?

JoanFM (dismissed):

cristianmtr (approved):

PR #3880, opened 2021-02-18
fix: fix and update docs and readmes for reciprocalrank

JoanFM:


evaluators/rank/ReciprocalRankEvaluator/manifest.yml
```diff
@@ -6,7 +6,7 @@ description: |
 author: Jina AI Dev-Team (dev-team@jina.ai)
 url: https://jina.ai
 vendor: Jina AI Limited
-documentation: https://github.com/jina-ai/jina-hub
+documentation: https://github.com/jina-ai/jina-hub/blob/master/evaluators/rank/NdcgEvaluator/README.md
 version: 0.0.5
```
bwanglzu, line 10: version needs to be updated?

bwanglzu (changes requested):

evaluators/rank/ReciprocalRankEvaluator/manifest.yml
JoanFM: good point, can u add the suggestion?

bwanglzu: @JoanFM I can't because it's not an edited line.

nan-wang: done

nan-wang (approved):
LGTM👍

theUnkownName (approved):

PR #4144, opened 2021-02-21
fix: fix bimatchranker

Yongxuanzhang:


rankers/BiMatchRanker/__init__.py
```diff
@@ -15,8 +15,8 @@ class BiMatchRanker(Chunk2DocRanker):
     D_MISS = 2000  # cost of a non-match chunk, used for normalization
 
     def _get_score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
-        s1 = self._directional_score(match_idx, match_chunk_meta, col=self.COL_MATCH_ID)
-        s2 = self._directional_score(match_idx, query_chunk_meta, col=self.COL_DOC_CHUNK_ID)
+        s1 = self._directional_score(match_idx, match_chunk_meta, col=self.COL_DOC_CHUNK_ID)
```
JoanFM, line 18: It also needs to change the required keys to 2 tuples ( one for query and one for match)

JoanFM (changes requested):

rankers/BiMatchRanker/__init__.py
Yongxuanzhang: Which line should I change? This one?
`required_keys = {'length'}`

PR #4775, opened 2021-02-28
fix: bimatchranker passing param

Yongxuanzhang:
Caused by https://github.com/jina-ai/jina/pull/1947
Several ways to solve this issue. Not sure which is the best. This issue caused the error of Tumblr-search.

1. See this PR
2. ` query_required_keys = ('length', )
       match_required_keys = ('length', )

    def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
        super().__init__(self.query_required_keys, self. match_required_keys, *args, **kwargs)`

### DIFF
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -59,7 +59,7 @@ class BiMatchRanker(Chunk2DocRanker):
     """
 
     def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
-        super().__init__(('length', ), ('length', ), *args, **kwargs)
+        super().__init__(query_required_keys=('length', ), match_required_keys=('length', ), *args, **kwargs)
         self.d_miss = d_miss or 2000
 
     def _get_score(self, match_idx: 'np.ndarray', query_chunk_meta: Dict, match_chunk_meta: Dict, *args, **kwargs):
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -51,6 +51,8 @@ class BiMatchRanker(Chunk2DocRanker):
 
     So for every `Document`, a score is computed adding some penalty to the `missing` in both perspectives.
 
+    :param query_required_keys: requested fields for query
+    :param match_required_keys: requested fields for match
     :param d_miss: Cost associated to a miss chunk
     :param args:  Additional positional arguments
     :param kwargs: Additional keyword arguments
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -58,9 +58,6 @@ class BiMatchRanker(Chunk2DocRanker):
     .. warning:: Here we suppose that the smaller chunk score means the more similar.
     """
 
-    # query_required_keys = ('length', )
-    # match_required_keys = ('length', )
-
     def __init__(self, query_required_keys: Optional = ('length',), match_required_keys: Optional = ('length',),
                  d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
         super().__init__(query_required_keys, match_required_keys, *args, **kwargs)
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -51,8 +51,6 @@ class BiMatchRanker(Chunk2DocRanker):
 
     So for every `Document`, a score is computed adding some penalty to the `missing` in both perspectives.
 
-    :param query_required_keys: requested fields for query
-    :param match_required_keys: requested fields for match
     :param d_miss: Cost associated to a miss chunk
     :param args:  Additional positional arguments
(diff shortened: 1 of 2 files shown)


rankers/BiMatchRanker/__init__.py
```diff
@@ -58,11 +57,13 @@ class BiMatchRanker(Chunk2DocRanker):
 
     .. warning:: Here we suppose that the smaller chunk score means the more similar.
     """
-    query_required_keys = ('length', )
-    match_required_keys = ('length', )
 
-    def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    # query_required_keys = ('length', )
```
JoanFM, line 61: remove these ones and it is good for me

JoanFM (changes requested):

rankers/BiMatchRanker/__init__.py
Yongxuanzhang: ✅

rankers/BiMatchRanker/__init__.py
```diff
@@ -58,11 +57,10 @@ class BiMatchRanker(Chunk2DocRanker):
 
     .. warning:: Here we suppose that the smaller chunk score means the more similar.
     """
-    query_required_keys = ('length', )
-    match_required_keys = ('length', )
 
-    def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, query_required_keys: Optional = ('length',), match_required_keys: Optional = ('length',),
+                 d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
```
bwanglzu, line 62: docstrings needs to be updated as well

bwanglzu (changes requested):

rankers/BiMatchRanker/__init__.py
Yongxuanzhang: 👍

rankers/BiMatchRanker/__init__.py
```diff
@@ -52,17 +51,18 @@ class BiMatchRanker(Chunk2DocRanker):
...
     :param kwargs: Additional keyword arguments
 
     .. warning:: Here we suppose that the smaller chunk score means the more similar.
     """
-    query_required_keys = ('length', )
-    match_required_keys = ('length', )
 
-    def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, query_required_keys: Optional = ('length',), match_required_keys: Optional = ('length',),
```
JoanFM, line 63: I think these should not be parameters, but they should be passed directly to the `super`constructor.

JoanFM (changes requested):

rankers/BiMatchRanker/__init__.py
Yongxuanzhang: In that case, the following one may be better? It depends on if we want to expose these params to users and the intention you refactored before,
`query_required_keys = ('length', )
match_required_keys = ('length', )

def init(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
super().init(self.query_required_keys, self. match_required_keys, *args, **kwargs)`

JoanFM: For now in this case, I think we do not want to expose it to users.

Yongxuanzhang: Sure, I will change to the second one. Or any other suggestions? 😄

JoanFM: The second one, and passed by value if possible I would say

Yongxuanzhang: Sure. Direct pass values now

rankers/BiMatchRanker/__init__.py
```diff
@@ -58,11 +57,9 @@ class BiMatchRanker(Chunk2DocRanker):
 
     .. warning:: Here we suppose that the smaller chunk score means the more similar.
     """
-    query_required_keys = ('length', )
-    match_required_keys = ('length', )
 
     def __init__(self, d_miss: Optional[Union[int, float]] = None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        super().__init__(('length', ), ('length', ), *args, **kwargs)
```
JoanFM, line 62: can u pass named arguments? like this if arguments change order in the future, this will still work

JoanFM (changes requested):

rankers/BiMatchRanker/__init__.py
Yongxuanzhang: Like this?
`super().__init__(query_required_keys=('length', ), match_required_keys=('length', ), *args, **kwargs)`

JoanFM (approved):

bwanglzu (approved):
LGTM!

Issue #3524, opened 2021-02-07
FAISS inner_product returns similarity and not distance


nan-wang:
this has an exception for metric inner_product, logic should be changed or warning added

_Originally posted by @JoanFM in https://github.com/jina-ai/jina-hub/pull/3057#discussion_r567464750_

JoanFM:
Solved by #4776

PR #4776, opened 2021-02-28
fix: faiss indexer inner_product returns sim and not distance (#3524)

friedsiberian:
Hi there,
in this fix results of query in a Faiss index will be always as top-k vector with smallest metric, so in ascending order.
The problem in this issue was specific on inner_product that returns similarity and not distances.

### DIFF
--- a/indexers/vector/FaissIndexer/__init__.py
+++ b/indexers/vector/FaissIndexer/__init__.py
@@ -130,6 +130,9 @@ class FaissIndexer(FaissDevice, BaseNumpyIndexer):
             from faiss import normalize_L2
             normalize_L2(vecs)
         dist, ids = self.query_handler.search(vecs, top_k)
+        if self.distance == 'inner_product':
+            self.logger.warning('inner_product will be output as distance.')
+            dist = np.abs(dist - 1)
         keys = self._int2ext_id[self.valid_indices][ids]
         return keys, dist
 
--- a/indexers/vector/FaissIndexer/__init__.py
+++ b/indexers/vector/FaissIndexer/__init__.py
@@ -82,6 +82,7 @@ class FaissIndexer(FaissDevice, BaseNumpyIndexer):
 
         metric = faiss.METRIC_L2
         if self.distance == 'inner_product':
+            self.logger.warning('inner_product will be output as distance instead of similarity.')
             metric = faiss.METRIC_INNER_PRODUCT
         if self.distance not in {'inner_product', 'l2'}:
             self.logger.warning('Invalid distance metric for Faiss index construction. Defaulting to l2 distance')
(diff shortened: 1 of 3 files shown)


indexers/vector/FaissIndexer/__init__.py
```diff
@@ -130,6 +130,9 @@ def query(self, vecs: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
             from faiss import normalize_L2
             normalize_L2(vecs)
         dist, ids = self.query_handler.search(vecs, top_k)
+        if self.distance == 'inner_product':
+            self.logger.warning('inner_product will be output as distance.')
```
maximilianwerk, line 134: This warning should rather go into the `build_advanced_index` method. Logging it with every request is overkill, in my opinion.

maximilianwerk (dismissed):
Could you please add another vector with different distances to clarify you point. If `inner_product` outputs similarity, shouldn't it be `1`, when a search with a vector returns the very same? So after your `1-dist` it should be `0` again, non?

indexers/vector/FaissIndexer/__init__.py
```diff
@@ -130,6 +130,9 @@ def query(self, vecs: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.nd
...
+            dist = np.abs(dist - 1)
```
JoanFM, line 135: is there a reason for this abs?

friedsiberian:
> Could you please add another vector with different distances to clarify you point. If `inner_product` outputs similarity, shouldn't it be `1`, when a search with a vector returns the very same? So after your `1-dist` it should be `0` again, non?

Hi @maximilianwerk, thanks for message.
My point was: `inner_product` originally returned similarity, so if query vector is equal to another vector in my index, `similarity` between them is 1. If I want to show distance between them, `distance = 1 - similarity` 
( np.abs(similarity - 1) is not needed )

I think I don't understand your questions.
Thanks

maximilianwerk:
> > Could you please add another vector with different distances to clarify you point. If `inner_product` outputs similarity, shouldn't it be `1`, when a search with a vector returns the very same? So after your `1-dist` it should be `0` again, non?
> 
> Hi @maximilianwerk, thanks for message.
> My point was: `inner_product` originally returned similarity, so if query vector is equal to another vector in my index, `similarity` between them is 1. If I want to show distance between them, `distance = 1 - similarity`
> ( np.abs(similarity - 1) is not needed )
> 
> What do you mean with these `Could you please add another vector with different distances to clarify you point.`?
> I think I don't understand your questions.
> Thanks

Sorry, my bad. When I read the code I mixed up the test behavior beforehand and afterwards. All good, thanks for clarifying this.

The second test is not needed, now that I fully understood, what happens.

JoanFM:
recheckcla

friedsiberian:
Hi, I have read the CLA Document and I hereby sign the CLA, thanks

JoanFM (approved):

Issue #5141, opened 2021-03-04
enable Sentencizer to segment text besides English


Yongxuanzhang:
The [Sentencizer](https://github.com/jina-ai/jina-hub/tree/master/segmenters/nlp/Sentencizer) only extracts and segments English words now. Because of this [line](https://github.com/jina-ai/jina-hub/blob/cb7a176535eed6c86973465e8c6bfc977a4c572e/segmenters/nlp/Sentencizer/__init__.py#L66).

We can modify this executor to make it applicable to other languages.

Need to add corresponding tests as well.

PR #5232, opened 2021-03-05
feat: fix sentencizer on other languages

Gracegrx:
Address issue https://github.com/jina-ai/jina-hub/issues/5141

Background:
Function `segment` of `Sentencizer` currently doesn't handle text in chinese and many other languages, because the filter uses string.printable, which is limited to english characters.

In this PR:
* Add another input argument to segment indicating which language the input text is, set default value as "en".
* Change segment to only apply filter on english texts.
* Add and update multiple tests for feature validation.

### DIFF
--- a/segmenters/nlp/Sentencizer/__init__.py
+++ b/segmenters/nlp/Sentencizer/__init__.py
@@ -47,11 +47,12 @@ class Sentencizer(BaseSegmenter):
                 self.min_sent_len, self.max_sent_len))
         self._slit_pat = re.compile('\s*([^{0}]+)(?<!\s)[{0}]*'.format(''.join(set(self.punct_chars))))
 
-    def segment(self, text: str, *args, **kwargs) -> List[Dict]:
+    def segment(self, text: str, lang="en", *args, **kwargs) -> List[Dict]:
         """
         Split the text into sentences.
 
         :param text: the raw text
+        :param lang: language of text, default is english
         :return: a list of chunk dicts with the split sentences
         :param args:  Additional positional arguments
         :param kwargs: Additional keyword arguments
@@ -63,7 +64,7 @@ class Sentencizer(BaseSegmenter):
         if not ret:
             ret = [(text, 0, len(text))]
         for ci, (r, s, e) in enumerate(ret):
-            f = ''.join(filter(lambda x: x in string.printable, r))
+            f = ''.join(filter(lambda x: x in string.printable, r)) if lang == "en" else r
             f = re.sub('\n+', ' ', f).strip()
             f = f[:self.max_sent_len]
             if len(f) > self.min_sent_len:
--- a/segmenters/nlp/Sentencizer/__init__.py
+++ b/segmenters/nlp/Sentencizer/__init__.py
@@ -22,6 +22,7 @@ class Sentencizer(BaseSegmenter):
         for example ['!', '.', '?'] will use '!', '.' and '?'
     :param uniform_weight: the definition of it should have
         uniform weight or should be calculated
+    :param lang: the language of text to be segmented
     :param args:  Additional positional arguments
     :param kwargs: Additional keyword arguments
 
@@ -32,6 +33,7 @@ class Sentencizer(BaseSegmenter):
                  max_sent_len: int = 512,
                  punct_chars: Optional[List[str]] = None,
                  uniform_weight: bool = True,
+                 lang: Optional[str] = 'en',
                  *args, **kwargs):
         """Set constructor."""
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


Yongxuanzhang (commented):
Need to bump the version. 😄

segmenters/nlp/Sentencizer/__init__.py
```diff
@@ -47,11 +47,12 @@ def __init__(self,
                 self.min_sent_len, self.max_sent_len))
         self._slit_pat = re.compile('\s*([^{0}]+)(?<!\s)[{0}]*'.format(''.join(set(self.punct_chars))))
 
-    def segment(self, text: str, *args, **kwargs) -> List[Dict]:
+    def segment(self, text: str, lang="en", *args, **kwargs) -> List[Dict]:
```
Yongxuanzhang, line 50: Would it be better to have this param in init()? e.g. `lang: str = 'en'`

segmenters/nlp/Sentencizer/tests/test_sentencizer.py
```diff
@@ -57,8 +57,38 @@ def validate(req):
...
+
+def test_sentencier_cn_lang():
+    """
+    If we indicate the text is not in english,
+    sentencizer should be able to skip string.printable and work for chinese.
+    """
+    sentencizer = Sentencizer()
+    text = '今天又是个大晴天！安迪回来以后，我们准备再去一次动物园！'
+    crafted_chunk_list = sentencizer.segment(text, "cn")
+    assert len(crafted_chunk_list) == 2
```
bwanglzu, line 78: I'd merge `test_sentencier_cn ` and `test_sentencier_cn_lang `, with

```python
@pytest.fixture
def chinese_sentence():
    return '今天是个大晴天！安迪回来以后，我们准备去动物园。'

@pytest.parameteriaze(
    'lang, expected',
    [('en', 0), ['cn', 2]],
)
def test_sentencier_cn(lang, expected, chinese_sentence):
    sentencizer = Sentencizer()
    segmented = sentencizer.segment(chinese_sentence, lang)
    assert len(segmented) == expected
```

Note, the above sample illustrates the idea, not tested.

Gracegrx: I tried combine multiple tests together, then found I wasn't able to use the fixture in parametrize... so in latest commit instead of using pytest.fixture, I made the two sentence global variables, please kindly provide suggestions if there's better way to do that, thanks!!

segmenters/nlp/Sentencizer/__init__.py
```diff
@@ -32,13 +33,15 @@ def __init__(self,
                  max_sent_len: int = 512,
                  punct_chars: Optional[List[str]] = None,
                  uniform_weight: bool = True,
+                 lang: Optional[str] = 'en',
```
florian-hoenicke, line 36: Please remove the Optional annotation. It is only used if the default is `None`

```suggestion
                 lang: str = 'en',

```

florian-hoenicke (changes requested):

segmenters/nlp/Sentencizer/__init__.py
JoanFM: There should be no language, the sentrncizer looks for characters regardless of the language.

JoanFM (changes requested):
There should be no language option, it does not behave differently for English than for other languages that support the same printable characters.

We can try to remove the check for printable characters and add tests on those languages with these characters.

segmenters/nlp/Sentencizer/__init__.py
florian-hoenicke: okay, makes sense.
`string.printable` only gives the following characters: `'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_{|}~ \t\n\r\x0b\x0c'`
So your suggestion would be to remove the filter step completely?
I would agree.

JoanFM: > Please remove the Optional annotation. It is only used if the default is `None`

It is used if the None option is properly handled, regardless of the default value.

JoanFM: I think we should try to remove and see how it reacts to different inputs, like Chinese characters, arabic ones, and maybe `accent marks` and umlauds from different languages

florian-hoenicke: You are right

segmenters/nlp/Sentencizer/tests/test_sentencizer.py
```diff
@@ -55,10 +56,28 @@ def validate(req):
...
     sentencizer = Sentencizer()
-    text = '今天是个大晴天！安迪回来以后，我们准备去动物园。'
-    crafted_chunk_list = sentencizer.segment(text, 0)
-    # Sentencizer does not work for chinese because string.printable does not contain Chinese characters
-    assert len(crafted_chunk_list) == 0
+    segmented = sentencizer.segment(sentence)
+    chunks = [i['text'] for i in segmented]
+    assert len(segmented) == expected_len
+    if (expected_len != 0):
+        assert chunks[0] == expected_text
```
florian-hoenicke, line 83: would be cool to validate against both chunks. Not just the first one

florian-hoenicke (approved):

JoanFM:
recheckcla

Gracegrx:

JoanFM (approved):

PR #5323, opened 2021-03-06
feat: increase verbosity for faiss indexer

rutujasurve94:
- Usage doc for `Faiss Indexer`


JoanFM (approved):
It would be nice if u could explain better the parameters

rutujasurve94:
> It would be nice if u could explain better the parameters

Adding param explanation in `README` as a subsequent PR, the `__init__.py` files explains the params.

PR #5505, opened 2021-03-08
feat: use single decorator for nlp crafters

JoanFM:


nan-wang:
@JoanFM Let's use the multistage Dockerfile to avoid installing pytest.

JoanFM:
> @JoanFM Let's use the multistage Dockerfile to avoid installing pytest.

can we do in another iteration?

nan-wang (approved):
LGTM👍

PR #5683, opened 2021-03-10
refactor:  fix bimatch ranker

Yongxuanzhang:
- Some of the tests are using the driver tests in Core. Since we moved group logic from ranker to driver, this Bimatchranker cannot be properly tested without the driver. 

- Meanwhile, the tests are not well organized, I can clean the code if the changes are ok. 
- The tests can pass on the latest master but for sure will fail on current CI, what should we do?

### DIFF
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -69,17 +69,27 @@ class BiMatchRanker(Chunk2DocRanker):
         s2 = self._directional_score(match_idx, query_chunk_meta, col=self.COL_QUERY_CHUNK_ID)
         return (s1 + s2) / 2.
 
+    def _group_by(self, match_idx, col_name):
+        # sort by ``col
+        _sorted_m = np.sort(match_idx, order=col_name)
+        _, _doc_counts = np.unique(_sorted_m[col_name], return_counts=True)
+        # group by ``col``
+        return np.split(_sorted_m, np.cumsum(_doc_counts))[:-1]
+
     def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
         # g [parent_match_id, match_id, query_id, score]
         # col = self.COL_DOC_CHUNK_ID, from matched_chunk aspect
         # col = self.COL_QUERY_CHUNK_ID, from query chunk aspect
         # group by "match_id" or "query_chunk_id". So here groups have in common 1st (match_parent_id) and `n` column
-
+        _groups = self._group_by(g, col)
+        # take the best match from each group
+        _groups_best = np.stack([np.sort(gg, order=col)[0] for gg in _groups])
+        # doc total length
         # how many chunks in the document (match or query)
-        _c = chunk_meta[g[0][col]]['length']
+        _c = chunk_meta[_groups_best[0][col]]['length']
         # how many chunks hit
-        _h = len(np.unique(g[col]))
+        _h = _groups_best.shape[0]
         # hit distance
-        sum_d_hit = np.sum(g[self.COL_SCORE])
-        # return score: all hit => 1, all_miss => 0
+        sum_d_hit = np.sum(_groups_best[self.COL_SCORE])
+        # all hit => 0, all_miss => 1
         return 1 - (sum_d_hit + self.d_miss * (_c - _h)) / (self.d_miss * _c)
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -74,15 +74,12 @@ class BiMatchRanker(Chunk2DocRanker):
         # col = self.COL_MATCH_ID, from matched_chunk aspect
         # col = self.COL_DOC_CHUNK_ID, from query chunk aspect
         # group by "match_id" or "query_chunk_id". So here groups have in common 1st (match_parent_id) and `n` column
-        _groups = self._group_by(g, col)
-        # take the best match from each group
-        _groups_best = np.stack([np.sort(gg, order=col)[0] for gg in _groups])
-        # doc total length
+
         # how many chunks in the document (match or query)
-        _c = chunk_meta[_groups_best[0][col]]['length']
+        _c = chunk_meta[g[0][col]]['length']
         # how many chunks hit
-        _h = _groups_best.shape[0]
+        _h = len(np.unique(g[col]))
         # hit distance
-        sum_d_hit = np.sum(_groups_best[self.COL_SCORE])
-        # all hit => 0, all_miss => 1
+        sum_d_hit = np.sum(g[self.COL_SCORE])
+        # return score: all hit => 1, all_miss => 0
         return 1 - (sum_d_hit + self.d_miss * (_c - _h)) / (self.d_miss * _c)
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -71,8 +71,8 @@ class BiMatchRanker(Chunk2DocRanker):
 
     def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
         # g [parent_match_id, match_id, query_id, score]
-        # col = self.COL_MATCH_ID, from matched_chunk aspect
-        # col = self.COL_DOC_CHUNK_ID, from query chunk aspect
+        # col = self.COL_DOC_CHUNK_ID, from matched_chunk aspect
+        # col = self.COL_QUERY_CHUNK_ID, from query chunk aspect
         # group by "match_id" or "query_chunk_id". So here groups have in common 1st (match_parent_id) and `n` column
 
         # how many chunks in the document (match or query)
--- a/rankers/BiMatchRanker/__init__.py
+++ b/rankers/BiMatchRanker/__init__.py
@@ -62,12 +62,12 @@ class BiMatchRanker(Chunk2DocRanker):
         super().__init__(query_required_keys=('length', ), match_required_keys=('length', ), *args, **kwargs)
         self.d_miss = d_miss or 2000
 
-    def _get_score(self, match_idx: 'np.ndarray', query_chunk_meta: Dict, match_chunk_meta: Dict, *args, **kwargs):
+    def score(self, match_idx: 'np.ndarray', query_chunk_meta: Dict, match_chunk_meta: Dict, *args, **kwargs):
         # match_idx [parent_match_id, match_id, query_id, score]
         # all matches have the same parent_id (the parent_id of the matches)
         s1 = self._directional_score(match_idx, match_chunk_meta, col=self.COL_DOC_CHUNK_ID)
         s2 = self._directional_score(match_idx, query_chunk_meta, col=self.COL_QUERY_CHUNK_ID)
-        return self.get_doc_id(match_idx), (s1 + s2) / 2.
+        return (s1 + s2) / 2.
 
     def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
         # g [parent_match_id, match_id, query_id, score]
--- a/rankers/BiMatchRanker/tests/test_bimatchranker.py
+++ b/rankers/BiMatchRanker/tests/test_bimatchranker.py
@@ -1,59 +1,108 @@
 import pytest
-import numpy as np
-
-from jina.executors.rankers import Chunk2DocRanker
+from jina import Document
+from jina.drivers.rank.aggregate import Chunk2DocRankDriver
+from jina.types.score import NamedScore
+from jina.types.sets import DocumentSet
 
 from .. import BiMatchRanker
 
+DISCOUNT_VAL = 0.5
+
+
+class SimpleChunk2DocRankDriver(Chunk2DocRankDriver):
+    def __init__(self, docs, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._docs = docs
+
+    @property
+    def exec_fn(self):
+        return self._exec_fn
+
+    @property
+    def docs(self):
+        return self._docs
+
+
+def create_document_to_score():
+    # doc: 1
+    # |- chunk: 2
+    # |  |- matches: (id: 4, parent_id: 40, score.value: 4),
+    # |  |- matches: (id: 5, parent_id: 50, score.value: 5),
+    # |
+    # |- chunk: 3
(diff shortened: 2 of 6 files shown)


rankers/BiMatchRanker/__init__.py
```diff
@@ -74,15 +74,12 @@ def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
...
-        # take the best match from each group
-        _groups_best = np.stack([np.sort(gg, order=col)[0] for gg in _groups])
-        # doc total length
+
         # how many chunks in the document (match or query)
-        _c = chunk_meta[_groups_best[0][col]]['length']
+        _c = chunk_meta[g[0][col]]['length']
         # how many chunks hit
-        _h = _groups_best.shape[0]
+        _h = len(np.unique(g[col]))
```
Yongxuanzhang, line 81: I have tested the _h and _c, the values are the same as before.

rankers/BiMatchRanker/__init__.py
```diff
@@ -74,15 +74,12 @@ def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
...
         # hit distance
-        sum_d_hit = np.sum(_groups_best[self.COL_SCORE])
-        # all hit => 0, all_miss => 1
+        sum_d_hit = np.sum(g[self.COL_SCORE])
```
Yongxuanzhang, line 83: This line I'm not sure.

nan-wang (dismissed):
Let's have a discussion about this PR.

rankers/LevenshteinRanker/tests/test_levenshteinranker.py
```diff
@@ -22,6 +16,6 @@ def test_levenshteinranker():
         match_meta
     )
 
-    assert len(new_scores) == 1
+    assert len(new_scores) == 2
```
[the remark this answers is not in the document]
Yongxuanzhang: Sure I will remove this

rankers/BiMatchRanker/__init__.py
```diff
@@ -74,15 +74,12 @@ def _directional_score(self, g: Dict, chunk_meta: Dict, col: str):
         # col = self.COL_MATCH_ID, from matched_chunk aspect
         # col = self.COL_DOC_CHUNK_ID, from query chunk aspect
```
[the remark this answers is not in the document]
Yongxuanzhang: 👍

Yongxuanzhang:
> Let's have a discussion about this PR.

We cannot use groupby in ranker since we have already moved it to driver

Yongxuanzhang:
Take the docstring as an example to explain:

    `Documents in the Index`:
        - Granularity 0 - Doc 1 - `BiMatchRanker is a Hub Executor. computes relevance scores. Found in Jina, the neural search framework`
                        - Granularity 1 - Doc 11 - `BiMatchRanker is a Ranking Jina Hub Executor`
                                        - Doc 12 - `It is a powerfull tool.`
                                        - Doc 13 - `Found in Jina`
                                        - Doc 14 - `the neural search framework`
        - Granularity 0 - Doc 2 - `A Ranker executor gives relevance Scores. Rankers are found in the Hub, and built automatically`
                        - Granularity 1 - Doc 21 - `A Ranker executor gives relevance Scores`
                        - Granularity 1 - Doc 22 - `Rankers are found in the Hub`
                        - Granularity 1 - Doc 23 - `and built automatically`

    `Query`
        - Granularity 0 - 'A Hub Executor. Ranking on Relevance scores. Giving automatically relevant documents from the Hub'
                        - Granularity 1 - `A Hub Executor`
                                        - `Ranking on Relevance scores`
                                        - `Giving automatically relevant documents from the Hub`

    `Semantic similarity matches` (TOP_K 3)
        Query chunk 1 -> `A Hub Executor` -> Doc11, Doc22, Doc21 (Children of `Doc1` and `Doc2`)
        Query chunk 2 -> `Ranking on Relevance scores` -> Doc21, Doc11, Doc22 (Children of `Doc2` and `Doc1`)
        Query chunk 3 -> `Scoring relevant documents`  -> Doc21, Doc22, Doc23 (Children of `Doc2`)

    In order to compute this relevance score, it tries to take into account two perspectives.

        - Match perspective: How many chunks of a match are part of the set of matches (hit) of a query (and its chunks).
            In the example:
                - `Doc1` has 4 chunks, but only 1 (hit) (Doc11) is found in the matches of the query chunks.
                - `Doc2` has 3 chunks and 2 of them (hit) are found in the matches (Doc21, Doc22).

        - Query perspective: Of all the chunks in the query, how many of them are matches by any of the chunks of each match document.
            In the example:
                - `Query` for `Doc1`: Query has 3 chunks, but only 2 of them are matched by chunks of match `Doc1` (Doc11)
                - `Query` for `Doc2`: Query has 3 chunks, and the 3 of them are matched by chunks of match `Doc2`

So we have 2 groups of chunks, one is from Doc1(2 chunks in total), another is from Doc2(7 chunks in total)

This is the expected results for _c and _h:
|             | _c when col = COL_DOC_CHUNK_ID | _h when col = COL_DOC_CHUNK_ID | _c when col = COL_QUERY_CHUNK_ID | _h when col = COL_QUERY_CHUNK_ID |
|-------------|--------------------------------|--------------------------------|----------------------------------|----------------------------------|
| Group1-Doc1 | 4                              | 1                              | 3                                | 2                                |
| Group2-Doc2 | 3                              | 3                              | 3                                | 3                                |

JoanFM (changes requested):
@Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test

Yongxuanzhang:
> @Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test

You can check the original code. Group_by is removed from ranker so there's no way to maintain the function

JoanFM:
> > @Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test
> 
> You can check the original code. Group_by is removed from ranker so there's no way to maintain the function

I see, anyway u just need to call the same function here, copying the code. But the group by is still needed in this specific case.

Yongxuanzhang:
> > > @Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test
> 
> > 
> 
> > You can check the original code. Group_by is removed from ranker so there's no way to maintain the function
> 
> 
> 
> I see, anyway u just need to call the same function here, copying the code. But the group by is still needed in this specific case.

But groupby is called in driver. So We shouldn't use that again in this ranker I think. It will do nothing. Because data is already grouped before calling this score().

I will also double check tomorrow on this.

JoanFM:
> > > > @Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test
> > 
> > 
> > > 
> > 
> > 
> > > You can check the original code. Group_by is removed from ranker so there's no way to maintain the function
> > 
> > 
> > I see, anyway u just need to call the same function here, copying the code. But the group by is still needed in this specific case.
> 
> But groupby is called in driver. So We shouldn't use that again in this ranker I think. It will do nothing. Because data is already grouped before calling this score().
> 
> You can check the calling in tests if possible.
> 
> I will also double check tomorrow on this.

Is a different grouping.

Yongxuanzhang:
> > > > > @Yongxuanzhang , the only change needed in _init_ is the name of the function, the functionality should not change. The only change is needed in the test
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > You can check the original code. Group_by is removed from ranker so there's no way to maintain the function
> > > 
> > > 
> > > I see, anyway u just need to call the same function here, copying the code. But the group by is still needed in this specific case.
> > 
> > 
> > But groupby is called in driver. So We shouldn't use that again in this ranker I think. It will do nothing. Because data is already grouped before calling this score().
> > You can check the calling in tests if possible.
> > I will also double check tomorrow on this.
> 
> Is a different grouping, it is confusing, but this specific executor does multiple groupings at different levels.
> 
> Herr instead of grouping by parent id, is grouping by chunks of the matches and chunks of the queries

Yes. You're correct. I think the new changes will get rid of using group_by and makes the process simpler?

Yongxuanzhang:
I'm still confused by this line, maybe we should sort by score instead of col? https://github.com/jina-ai/jina-hub/blob/b14a61263a048d18fcbd8ffa4cbe2a6b5f85c5e5/rankers/BiMatchRanker/__init__.py#L79

JoanFM:
> I'm still confused by this line, maybe we should sort by score instead of col? Why do we think the chunks with smallest col is the best? @JoanFM
> 
> https://github.com/jina-ai/jina-hub/blob/b14a61263a048d18fcbd8ffa4cbe2a6b5f85c5e5/rankers/BiMatchRanker/__init__.py#L79

to be honest I am not sure, but lets first refactor the test keeping things as they are, then u can face a refactor directly

Yongxuanzhang:
> > I'm still confused by this line, maybe we should sort by score instead of col? Why do we think the chunks with smallest col is the best? @JoanFM
> > https://github.com/jina-ai/jina-hub/blob/b14a61263a048d18fcbd8ffa4cbe2a6b5f85c5e5/rankers/BiMatchRanker/__init__.py#L79
> 
> to be honest I am not sure, but lets first refactor the test keeping things as they are, then u can face a refactor directly

Sure! I will roll back all the changes!

rankers/BiMatchRanker/tests/test_bimatchranker.py
```diff
@@ -137,9 +137,9 @@ def test_bimatchranker_readme():
 
     assert match_score_1 > match_score_2
     assert doc.matches[0].id == '2'
-    assert match_score_1 == pytest.approx(0.9994, 0.001)
+    assert match_score_1 == pytest.approx(0.5333, 0.001)
```
JoanFM, line 140: why does the value change? it should not be changed by a refactor

JoanFM (changes requested):

rankers/BiMatchRanker/tests/test_bimatchranker.py
Yongxuanzhang: This is the original value. This can be change with d_miss setting

Yongxuanzhang: https://github.com/jina-ai/jina-hub/blob/b14a61263a048d18fcbd8ffa4cbe2a6b5f85c5e5/rankers/BiMatchRanker/tests/test_bimatchranker.py#L77
https://github.com/jina-ai/jina-hub/blob/b14a61263a048d18fcbd8ffa4cbe2a6b5f85c5e5/rankers/BiMatchRanker/tests/test_bimatchranker.py#L84

Yongxuanzhang: Now this is the same as before

Yongxuanzhang:
I reorganized the tests so we can better compare the current tests with previous ones. @JoanFM

rankers/BiMatchRanker/tests/test_bimatchranker.py
```diff
@@ -105,6 +43,38 @@ def create_data(query_chunk2match_chunk):
...
+        ],
+        110: [
+            {'parent_id': 1, 'id': 10, 'score': 0.3, 'length': 200},
+            {'parent_id': 1, 'id': 11, 'score': 0.2, 'length': 200},
+            {'parent_id': 4294967294, 'id': 20, 'score': 0.1, 'length': 300},
+        ]
+    }
+
+    doc = create_data(query_chunk2match_chunk)
+    driver = SimpleChunk2DocRankDriver(
```
JoanFM, line 59: is it possible to avoid using the driver here? we try to avoid depending from hub to the core.

JoanFM (changes requested):

rankers/BiMatchRanker/tests/test_bimatchranker.py
Yongxuanzhang: I feel like introducing driver with keep the same tests behaviour as before

JoanFM: but it adds a dependency that is unwanted, can u mimic the behavior urself?

theUnkownName: So are we going to test then one scalar instead of the returned list? I'm also not sure how to do the other tests

theUnkownName: Ahhh sorry just saw last Joans comment. To make sure, we want to mimic getting the list without using a driver, right?

Yongxuanzhang: And we lack integration tests for hub executors, test on a higher level will help us find out more potential bugs. Like the required_keys before. We both pass the tests in core and hub but still failed the examples.

Yongxuanzhang: Ok I see, that's a good argument

JoanFM:
@Yongxuanzhang u can direct the PR directly to master

rankers/SimpleAggregateRanker/__init__.py
```diff
@@ -38,4 +38,4 @@ def score(self, match_idx, query_chunk_meta, match_chunk_meta, *args, **kwargs):
                 raise ValueError(f'Setting "is_reversed_score" to True does not allow to have an aggregated document '
                                  f'negative score')
             aggregated_score = 1. / (1. + aggregated_score)
-        return aggregated_score
+        return self.get_doc_id(match_idx), aggregated_score
```
JoanFM: rankers do not return ids anymore

JoanFM (changes requested):

rankers/SimpleAggregateRanker/__init__.py
Yongxuanzhang: I'm getting rid of other changes and only keep the Bimatchranker. 😂 
That's why I propose to merge into your rather than master.

JoanFM: as u want, but u can target master

JoanFM: so just copy the master folder from SimpleAggregateRanker and handle that ranker in a different branch

JoanFM (approved):

Issue #5142, opened 2021-03-04
provide more features of JiebaSegmenter


Yongxuanzhang:
We can provide more choices for users to enable them to use more functions of the Jieba library by exposing param to constructor.
Like jieba.load_userdict().

Needs to have tests for any new features.

nan-wang:
Yes, this would be helpful

PR #5691, opened 2021-03-10
feat(segmenter): add user dict to JiebaSegmenter

jacobowitz:
This PR adds a new option to the `JiebaSegmenter` by exposing an existing feature in the Jieba library:

Allow the user to extend the default dict by a custom dict of their choice. The dict needs to be specified as a path in the `user_dict_file` parameter. The format of the dict is specified in the [Jieba repository](https://github.com/fxsjy/jieba#load-dictionary)

This feature was requested in this [issue](https://github.com/jina-ai/jina-hub/issues/5142), which did not specify exactly which features to expose. 
Potentially one could also expose the functionality to [modify the existing dict](https://github.com/fxsjy/jieba#modify-dictionary). That seemed a bit too cumbersome for the user to me though.
The custom dict feature assumes that the dictionary file is accessible locally. If that requirement is too inconvenient, one could also try to get it via HTTP or so.

All changes are covered with unit tests.

### DIFF
--- a/segmenters/nlp/JiebaSegmenter/__init__.py
+++ b/segmenters/nlp/JiebaSegmenter/__init__.py
@@ -20,7 +20,6 @@ class JiebaSegmenter(BaseSegmenter):
     :type user_dict_file: str
     :raises:
         ValueError: If `mode` is not any of the expected modes
-        FileNotFoundError: If `user_dict_file` does not exist
     :param args:  Additional positional arguments
     :param kwargs: Additional keyword arguments
 
@@ -28,18 +27,22 @@ class JiebaSegmenter(BaseSegmenter):
 
     def __init__(self, mode: str = 'accurate', user_dict_file: str = None, *args, **kwargs):
         """Set Constructor."""
-        import jieba
-        import os
-
         super().__init__(*args, **kwargs)
         if mode not in ('accurate', 'all', 'search'):
             raise ValueError('you must choose one of modes to cut the text: accurate, all, search.')
         self.mode = mode
-
-        if user_dict_file is not None:
-            if not os.path.exists(user_dict_file):
-                raise FileNotFoundError(f'User dictionary can not be found at {user_dict_file}')
         self.user_dict_file = user_dict_file
+
+    def post_init(self):
+        """ Load custom dict if provided. Raise FileNotFoundError if dict does not exist."""
+        super().post_init()
+
+        if self.user_dict_file is not None:
+            import os
+            import jieba
+            if not os.path.exists(self.user_dict_file):
+                raise FileNotFoundError(f'User dictionary can not be found at {self.user_dict_file}')
+            jieba.load_userdict(self.user_dict_file)
 
     def segment(self, text: str, *args, **kwargs) -> List[Dict]:
         """
@@ -53,9 +56,6 @@ class JiebaSegmenter(BaseSegmenter):
         :rtype: List[Dict]
         """
         import jieba
-
-        if self.user_dict_file is not None:
-            jieba.load_userdict(self.user_dict_file)
 
         if self.mode == 'search':
             words = jieba.cut_for_search(text)
--- a/segmenters/nlp/JiebaSegmenter/__init__.py
+++ b/segmenters/nlp/JiebaSegmenter/__init__.py
@@ -56,7 +56,6 @@ class JiebaSegmenter(BaseSegmenter):
         :rtype: List[Dict]
         """
         import jieba
-
         if self.mode == 'search':
             words = jieba.cut_for_search(text)
         elif self.mode == 'all':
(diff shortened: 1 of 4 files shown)


segmenters/nlp/JiebaSegmenter/__init__.py
```diff
@@ -14,20 +14,43 @@ class JiebaSegmenter(BaseSegmenter):
      - all
 
     :type mode: str
+    :param pool_size: Number of parallel processes to use.
```
JoanFM, line 17: Hello @jacobowitz ,

Thank you very much for the proposal. I would however not expose the multiprocessing at this level. Tipically parallelization in Jina is handled at Pod level by having multiple Peas inside a single Pod.

This could help at improving the throughput of a single request, but it may not play well with the current parallelization model in some cases. So at least parallelization should be disabled by default

segmenters/nlp/JiebaSegmenter/__init__.py
```diff
@@ -14,20 +14,43 @@ class JiebaSegmenter(BaseSegmenter):
...
+        FileNotFoundError: If `user_dict_file` does not exist
     :param args:  Additional positional arguments
     :param kwargs: Additional keyword arguments
 
     """
 
-    def __init__(self, mode: str = 'accurate', *args, **kwargs):
+    def __init__(self, mode: str = 'accurate', pool_size: int = 1, user_dict_file: str = None, *args, **kwargs):
         """Set Constructor."""
+        from multiprocessing import cpu_count
```
JoanFM, line 35: Can you keep the imports in post_init? This way we can avoid some warnings when using jina

JoanFM (changes requested):

segmenters/nlp/JiebaSegmenter/__init__.py
jacobowitz: It is disabled by default as `pool_size` is set to one. But I was not aware of the Pod level handling of parallelization, which sounds indeed like better idea. Setting this to some high number here could potentially mess up resource management there I guess.
So let me remove the pool_size option quickly, will also address your other comment then and my failing test :)

segmenters/nlp/JiebaSegmenter/__init__.py
```diff
@@ -14,20 +14,33 @@ class JiebaSegmenter(BaseSegmenter):
...
+        FileNotFoundError: If `user_dict_file` does not exist
     :param args:  Additional positional arguments
     :param kwargs: Additional keyword arguments
 
     """
 
-    def __init__(self, mode: str = 'accurate', *args, **kwargs):
+    def __init__(self, mode: str = 'accurate', user_dict_file: str = None, *args, **kwargs):
         """Set Constructor."""
+        import jieba
```
JoanFM, line 31: no need to add jieba here as dependency

segmenters/nlp/JiebaSegmenter/__init__.py
```diff
@@ -40,6 +53,10 @@ def segment(self, text: str, *args, **kwargs) -> List[Dict]:
         :rtype: List[Dict]
         """
         import jieba
```
Yongxuanzhang, line 55: since we import jieba in init, I wonder if we need to import it again?

segmenters/nlp/JiebaSegmenter/__init__.py
```diff
@@ -40,6 +53,10 @@ def segment(self, text: str, *args, **kwargs) -> List[Dict]:
...
+
+        if self.user_dict_file is not None:
```
JoanFM, line 57: This part and the checking of the existance of user dict file should be handled at `post_init()` method.

Post init is a special method in Jina where dependencies are loaded and expensive setting is done. It is our common pattern to work like this.

This way you do not need to load the user dict for every segment call

JoanFM (changes requested):

segmenters/nlp/JiebaSegmenter/tests/test_jiebasegmenter.py
```diff
@@ -1,8 +1,33 @@
...
     jieba_crafter = JiebaSegmenter(mode='accurate')
     text = '今天是个大晴天！安迪回来以后，我们准备去动物园。'
     crafted_chunk_list = jieba_crafter.segment(text, 0)
     assert len(crafted_chunk_list) == 14
+
+
+def test_jieba_user_dir():
+    jieba_segmenter = JiebaSegmenter()
+    text = '今天是个大晴天！安迪回来以后，我们准备去动物园。thisisnotachineseword'
+    crafted_chunk_list = jieba_segmenter.segment(text, 0)
```
Yongxuanzhang, line 20: No need to pass 0 here I think, could u try to remove it?

segmenters/nlp/JiebaSegmenter/__init__.py
jacobowitz: ah that sounds better indeed, thx.
I am using post_init now and also call super there as I saw it in other places, but I am not 100% sure if that is correct here. I suppose so?

JoanFM: Yes, it is great!

jacobowitz:

Yongxuanzhang:
Cool! Thank you @jacobowitz

PR #6013, opened 2021-03-14
Feat nltk sentencizer 5790

mezig351:
My implementation for Sentencizer with NLTK

### DIFF
--- a/segmenters/nlp/NLTKSentencizer/__init__.py
+++ b/segmenters/nlp/NLTKSentencizer/__init__.py
@@ -1,6 +1,6 @@
 from typing import Dict, List, Optional
-
 from jina.executors.segmenters import BaseSegmenter
+from jina.executors.decorators import single
 
 
 class NLTKSentencizer(BaseSegmenter):
@@ -97,6 +97,7 @@ class NLTKSentencizer(BaseSegmenter):
                 raise
         self.logger.info(f'NLTK {self.language} sentence tokenizer ready.')
 
+    @single
     def segment(self, text: str, *args, **kwargs) -> List[Dict]:
         """
         Segment text into sentences.
--- a/segmenters/nlp/NLTKSentencizer/__init__.py
+++ b/segmenters/nlp/NLTKSentencizer/__init__.py
@@ -1,0 +1,109 @@
+from typing import Dict, List
+
+from jina.executors.segmenters import BaseSegmenter
+
+
+class NLTKSentencizer(BaseSegmenter):
+    """
+    Segment text into sentences using :class:`nltk.PunktSentenceTokenizer`
+    for the specified language.
+
+    Example:
+
+    >>> sentencizer = NLTKSentencizer() # doctest:+ELLIPSIS
+    NLTKSentencizer@...[I]:post_init may take some time...
+    NLTKSentencizer@...[I]:NLTK english sentence tokenizer ready.
+    NLTKSentencizer@...[I]:post_init may take some time takes ... seconds (...s)
+    >>> text = "Today is a good day. Can't wait for tomorrow!"
+    >>> sentencizer.segment(text)
+    [{'text': 'Today is a good day.', 'offset': 0, 'location': [0, 20]}, {'text': "Can't wait for tomorrow!", 'offset': 1, 'location': [21, 45]}]
+
+
+    :param language: default='english'. Lowercased language name to initialize the sentence tokenizer, accepted languages are listed in :attr:`SUPPORTED_LANGUAGES`.
+    :type language: str
+    :param args: Additional positional arguments
+    :param kwargs: Additional keyword arguments
+    """
+
+    SUPPORTED_LANGUAGES = [
+        'czech',
+        'danish',
+        'dutch',
+        'english',
+        'estonian',
+        'finnish',
+        'french',
+        'german',
+        'greek',
+        'italian',
+        'norwegian',
+        'polish',
+        'portuguese',
+        'russian',
+        'slovene',
+        'spanish',
+        'swedish',
+        'turkish',
+    ]
+
+    def __init__(self, language: str = 'english', *args, **kwargs):
+        """Set constructor"""
+        super().__init__(*args, **kwargs)
+        self.language = language
+        self.sent_tokenizer = None
+
+    def post_init(self):
+        from nltk.data import load
+
+        try:
+            self.sent_tokenizer = load(f'tokenizers/punkt/{self.language}.pickle')
+            self.logger.info(f'NLTK {self.language} sentence tokenizer ready.')
+        except LookupError:
+            if self.language in self.SUPPORTED_LANGUAGES:
+                try:
+                    import nltk
+
+                    nltk.download('punkt')
+                    self.post_init()
+                except:
+                    self.logger.error(
+                        f'Please ensure that "nltk_data" folder is accessible to your working directory.'
+                    )
+                    raise
+            else:
+                self.logger.error(
+                    f'The language you specified ("{self.language}") is not supported by NLTK. '
+                    f'Please ensure that language is one of the acceptable languages listed in '
+                    f'{self.SUPPORTED_LANGUAGES}.\nOr for latest list of supported languages check out '
+                    f'https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/index.xml with '
+                    f'id="punkt"'
+                )
+                raise
+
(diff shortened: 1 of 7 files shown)


segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        'spanish',
+        'swedish',
+        'turkish',
+    ]
+
+    def __init__(self, language: str = 'english', *args, **kwargs):
+        """Set constructor"""
+        super().__init__(*args, **kwargs)
+        self.language = language
+        self.sent_tokenizer = None
```
JoanFM, line 53: Avoid setting `sent_tokenizer`. This is something very Jina specific to avoid serializing expensive models i

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            self.sent_tokenizer = load(f'tokenizers/punkt/{self.language}.pickle')
+            self.logger.info(f'NLTK {self.language} sentence tokenizer ready.')
+        except LookupError:
+            if self.language in self.SUPPORTED_LANGUAGES:
+                try:
+                    import nltk
+
+                    nltk.download('punkt')
+                    self.post_init()
```
JoanFM, line 67: this seems like a dangerous recursive call

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        """Set constructor"""
+        super().__init__(*args, **kwargs)
+        self.language = language
+        self.sent_tokenizer = None
+
+    def post_init(self):
+        from nltk.data import load
+
+        try:
+            self.sent_tokenizer = load(f'tokenizers/punkt/{self.language}.pickle')
```
JoanFM, line 59: why do not we pass a `tokenizer_path` as parameter to the object, from where to load the tokenizer object?

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    [{'text': 'Today is a good day.', 'offset': 0, 'location': [0, 20]}, {'text': "Can't wait for tomorrow!", 'offset': 1, 'location': [21, 45]}]
+
+
+    :param language: default='english'. Lowercased language name to initialize the sentence tokenizer, accepted languages are listed in :attr:`SUPPORTED_LANGUAGES`.
+    :type language: str
+    :param args: Additional positional arguments
+    :param kwargs: Additional keyword arguments
+    """
+
+    SUPPORTED_LANGUAGES = [
```
JoanFM, line 28: is it a need to keep this exhaustive list?

JoanFM: doesn't NLTK have a list of supported languages in their API?

segmenters/nlp/NLTKSentencizer/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jina
```
JoanFM, line 1: please remove `jina` as dependency, since the Docker image is built with jina images as base

segmenters/nlp/NLTKSentencizer/requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+nltk
```
JoanFM, line 2: can u set a version?

JoanFM (changes requested):

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -50,21 +50,20 @@ def __init__(self, language: str = 'english', *args, **kwargs):
         """Set constructor"""
         super().__init__(*args, **kwargs)
         self.language = language
-        self.sent_tokenizer = None
+        self.tokenizer_path = f'tokenizers/punkt/{self.language}.pickle'
```
JoanFM, line 53: I am not sure, but is this a default parametrer?

segmenters/nlp/NLTKSentencizer/__init__.py
mezig351: I added tokenizer_path instead of sent_tokenizer to init, but I did not make it a parameter to the object. Because the nltk.data.load function only searches under nltk_data/tokenizers/punkt; otherwise throws an error. And path of nltk_data has to be one of the accepted paths, default being ~/nltk_data . So, it felt like it would not really be a parameter. Furthermore, the saved pickle file has to be of correct class etc. to load properly. 
Do you think it's ok to keep it like this?

JoanFM: so u mean, the nltk will not accept an input path from another path? Why do they accept it as a parameter then?

Would it make sense to have it as Optional parameter  that defaults to the value you claim if not present?

segmenters/nlp/NLTKSentencizer/__init__.py
mezig351: Their API is very unspecific about the supported languages.. I had added this to make errors more understandable.  The punkt tokenizer models are being downloaded from here: https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/index.xml , so I guess one way could be writing an XML parser to read and load this instead of keeping fixed names of languages.

But as I said, the main point of keeping this list was to make the error more understandable. It throws the same error if nltk_data folder is not found, if a language name is mis-spelt, or a non-existing model for a language is called. 

If you think it's more appropriate to delete it, I can definitely do that.

segmenters/nlp/NLTKSentencizer/__init__.py
mezig351: Yeah, you were right. I checked the load function, and default protocol searches that folder path but it's possible to specify with "file://..." . I am making the changes 👍

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -46,38 +48,49 @@ class NLTKSentencizer(BaseSegmenter):
         'turkish',
     ]
 
-    def __init__(self, language: str = 'english', *args, **kwargs):
+    def __init__(
+        self, language: str = 'english', resource_url: str = None, *args, **kwargs
```
JoanFM, line 52: if it is Optional, let's have a proper `type` annotation with `Optional[str]`

JoanFM: Also `language` could be set to `Optional` and use `None` instead of empty string to detect if is provided

JoanFM (changes requested):

JoanFM:
recheckcla

mezig351:
**_I have read the CLA Document and I hereby sign the CLA_**

segmenters/nlp/NLTKSentencizer/manifest.yml
```diff
@@ -0,0 +1,13 @@
+manifest_version: 1
+name: NLTKSentencizer
+kind: segmenter
+description:
+  Segments text into sentences using NLTK.
+author: Jina AI Dev-Team (dev-team@jina.ai)
```
nan-wang, line 6:
```suggestion
author: mezig351 (your@email.com)
```

nan-wang (changes requested):
LGTM

nan-wang (dismissed):
LGTM👍

segmenters/nlp/NLTKSentencizer/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+                )
+                raise
+            else:
+                self.logger.error(
+                    f'The resource_url you provided could not be loaded: {self.resource_url}'
+                )
+                raise
+        self.logger.info(f'NLTK {self.language} sentence tokenizer ready.')
+
+    def segment(self, text: str, *args, **kwargs) -> List[Dict]:
```
JoanFM, line 100: Hey @mezig351 , since the last patch, we need the `segmenters` to be decorated with `single` decorator to make sure it works with the rest of the pipeline. It is not yet documented, but it is needed, you can see some `segmenters` that have been recently merged into master to see how that works

JoanFM (changes requested):

JoanFM (approved):

JoanFM:
Thank you very much for your contribution @mezig351 , really nice feature!

PR #6373, opened 2021-03-18
feat: add FaceNetEncoder

strawberrypie:
This PR adds FaceNet encoder to be used in Jina. 

There is a legacy [code example](https://github.com/jina-ai/legacy-examples/tree/main/face-db-search/Encoder) that also uses FaceNet, but it incorrectly assumes that the entire image is a face. This encoder works differently:
* Face detector detects faces on the image
  * If multiple faces are detected the largest is selected (default heuristic)
  * If no faces are detected a dummy face is used (an array with zeroes)
* The face is encoded to an embedding


JoanFM:
Hey @strawberrypie,

Thank you for your contribution.

However, as per how you describe the feature, I think this executor should be divided into different ones.

I think there should be a `segmenter` for `detecting faces` (similar to `TorchObjectDetectionSegmenter`) and then have the encoder itself where it is assumed that the faces are there.

We try to keep the granularity of the tasks for maximizing flexibility, these 2 executors could easily be used in combination using a `CompoundExecutor`.

JoanFM (changes requested):
I would divide the concerns of detection and embedding into different `executors`

JoanFM:
> Thanks for clarification, it makes sense to have separate face detector for other use cases. Should I split this PR into two (segmenter + encoder)? @JoanFM

That would be really helpful !

PR #6375, opened 2021-03-18
fix(bimatch): fix group by order

Yongxuanzhang:
`_groups_best = np.stack([np.sort(gg, order=col)[0] for gg in _groups])`
should sort by 'score' to get the best, this won't have effect if we use default d_miss value since it has a large value.


nan-wang (approved):
LGTM👍

PR #6654, opened 2021-03-21
feat: add FaceNetSegmenter

strawberrypie:
This PR adds FaceNetSegmenter to be used in Jina. This class detects faces on an image and returns them as an NumPy array.  
This is a continuation of work being done in #6373.

Note that this segmenter could support batching, but other segmenters do not do that. I believe that it needs to be done in a separate PR.

PR #6655, opened 2021-03-21
feat: add FaceNetEncoder

strawberrypie:
This PR adds FaceNetEncoder to be used in Jina. This class encodes face images. Works best with images from FaceNetSegmenter (#6654), as they are intended to work together.
This is a continuation of work being done in #6373.


encoders/image/FaceNetEncoder/tests/test_facenetencoder.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.fixture
+def encoder():
+    return FaceNetEncoder()
+
+
+@pytest.mark.parametrize('batch_size', [1, 2, 32])
+def test_encoding_face_batch(encoder, batch_size: int):
+    image = Image.open(ENCODER_DIR / 'imgs' / 'man_piercing.jpg')
```
nan-wang, line 19:
```suggestion
    import os
    image = Image.open(os.path.join(ENCODER_DIR, 'imgs', 'man_piercing.jpg'))
```

encoders/image/FaceNetEncoder/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self,
+                 pretrained_weights: str = 'vggface2',
+                 channel_axis: int = 1,
+                 *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.pretrained_weights = pretrained_weights
+        self.channel_axis = channel_axis
+
+        self.embedder = None
```
nan-wang, line 33: Claiming the `embedder` in `__init__` will lead to storing it when dumping. This is not what we want because the model can always be loaded on the spot. BTW, what about using `model` instead of `embedder`?
```suggestion
```

encoders/image/FaceNetEncoder/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+        :param kwargs: Additional positional arguments.
+        :return: A `BatchSize x EmbeddingDimension` numpy array.
+        """
+        if self.channel_axis != self._default_channel_axis:
+            data = np.moveaxis(data, self.channel_axis, self._default_channel_axis)
+
+        with torch.no_grad():
+            images = torch.from_numpy(data.astype('float32')).to(self.device)
+            embedded_faces = self.embedder(images)
+        return embedded_faces.cpu()
```
nan-wang, line 62:
```suggestion
        return embedded_faces.detach().numpy()
```

nan-wang (changes requested):

nan-wang:
Great contribution!

encoders/image/FaceNetEncoder/__init__.py
strawberrypie: Sure, I wasn't familiar with serialization logic. pickle would serialize this property in `BaseExecutor.save` if it is defined in `__init__`.

nan-wang (approved):
LGTM👍

PR #7357, opened 2021-04-01
fix: fix TransformerTorchEncoder to use RetriBert models

JoanFM:
Some family of models from `transformers` do not seem to align the common `BaseModelOutput` pattern, and they need to extract embeddings in a different way.


maximilianwerk (approved):
LGTM

PR #7359, opened 2021-04-01
feat: added huggingface api calls

maximilianwerk:
Enables the usage of the huggingface API instead of calling a model locally. To enable it, just add an API token to the encoder. The model will not be loaded locally at all.

```
!TransformerTorchEncoder
with:
  pooling_strategy: auto
  pretrained_model_name_or_path: distilbert-base-cased
  max_length: 192
  api_token: ${{JINA_HUGGINGFACE_API_TOKEN}}
```

### DIFF
--- a/encoders/nlp/TransformerTorchEncoder/__init__.py
+++ b/encoders/nlp/TransformerTorchEncoder/__init__.py
@@ -143,11 +143,11 @@ class TransformerTorchEncoder(TorchDevice, BaseEncoder):
         import requests
 
         retries = 0
+        headers = {"Authorization": "Bearer " + self.api_token}
+        api_url = f"https://api-inference.huggingface.co/pipeline/feature-extraction/{self.pretrained_model_name_or_path}"
 
         while retries < self.max_retries:
             retries += 1
-            headers = {"Authorization": "Bearer " + self.api_token}
-            api_url = f"https://api-inference.huggingface.co/pipeline/feature-extraction/{self.pretrained_model_name_or_path}"
             r = requests.post(api_url, json=query, headers=headers)
             if r.status_code == HTTP_SERVICE_UNAVAILABLE:
                 self.logger.info('Service is currently unavailable. Model is loading.')
--- a/encoders/nlp/TransformerTorchEncoder/__init__.py
+++ b/encoders/nlp/TransformerTorchEncoder/__init__.py
@@ -237,7 +237,4 @@ class TransformerTorchEncoder(TorchDevice, BaseEncoder):
                         return outputs.cpu().numpy()
                     hidden_states = outputs.hidden_states
 
-            if not isinstance(outputs, torch.Tensor):
-                return self._compute_embedding(hidden_states, input_tokens)
-            else:
-                return outputs.cpu().numpy()
+            return self._compute_embedding(hidden_states, input_tokens)
--- a/encoders/nlp/TransformerTorchEncoder/__init__.py
+++ b/encoders/nlp/TransformerTorchEncoder/__init__.py
@@ -6,14 +6,12 @@ import time
 from typing import Optional, Dict
 
 import numpy as np
-import requests
 
 from jina.executors.decorators import batching, as_ndarray
 from jina.executors.devices import TorchDevice
 from jina.executors.encoders import BaseEncoder
 
 if False:
-    # It is not assumed yet they inherit from this, but the transformers documentation seem to suggest so
     import torch
 
 HTTP_SERVICE_UNAVAILABLE = 503
@@ -128,6 +126,8 @@ class TransformerTorchEncoder(TorchDevice, BaseEncoder):
                 self.model = torch.quantization.quantize_dynamic(
                     self.model, {torch.nn.Linear}, dtype=torch.qint8
                 )
+        else:
+            self._api_call('Scotty, please warmup.')
 
     def amp_accelerate(self):
         """Check acceleration method """
@@ -140,6 +140,8 @@ class TransformerTorchEncoder(TorchDevice, BaseEncoder):
             return nullcontext()
 
     def _api_call(self, query):
+        import requests
+
         retries = 0
 
         while retries < self.max_retries:
--- a/encoders/nlp/TransformerTorchEncoder/__init__.py
+++ b/encoders/nlp/TransformerTorchEncoder/__init__.py
@@ -1,10 +1,12 @@
 __copyright__ = "Copyright (c) 2021 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-import os
+import time
+
 from typing import Optional, Dict
 
 import numpy as np
+import requests
 
 from jina.executors.decorators import batching, as_ndarray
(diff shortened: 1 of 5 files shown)


encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -1,18 +1,22 @@
 __copyright__ = "Copyright (c) 2021 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
-import os
+import time
+
 from typing import Optional, Dict
 
 import numpy as np
+import requests
```
JoanFM, line 9: hide this import inside the executor code

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -120,9 +139,25 @@ def amp_accelerate(self):
...
+        retries = 0
+
+        while retries < self.max_retries:
+            retries += 1
+            headers = {"Authorization": "Bearer " + self.api_token}
+            api_url = f"https://api-inference.huggingface.co/pipeline/feature-extraction/{self.pretrained_model_name_or_path}"
+            r = requests.post(api_url, json=query, headers=headers)
+            if r.status_code == HTTP_SERVICE_UNAVAILABLE:
+                self.logger.info('Service is currently unavailable. Model is loading.')
+                time.sleep(3)
```
JoanFM, line 152: this is very long right? If the cause for this is having the model loading? should we send a DUMMY call at `post_init` to have it warm?

So later, during proper encoding, we do not wait so much

encoders/nlp/TransformerTorchEncoder/__init__.py
JoanFM: to avoid warnings

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -1,18 +1,22 @@
...
 
 import numpy as np
+import requests
 
 from jina.executors.decorators import batching, as_ndarray
 from jina.executors.devices import TorchDevice
 from jina.executors.encoders import BaseEncoder
 
 if False:
     # It is not assumed yet they inherit from this, but the transformers documentation seem to suggest so
```
JoanFM, line 16: remove this comment, as it does not make sense anymore

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/__init__.py
maximilianwerk: Yes :+1:

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -120,9 +139,25 @@ def amp_accelerate(self):
...
             return nullcontext()
 
-    def _compute_embedding_from_model_output(self, outputs: 'BaseModelOutput', input_tokens: Dict):
+    def _api_call(self, query):
+        retries = 0
+
+        while retries < self.max_retries:
+            retries += 1
+            headers = {"Authorization": "Bearer " + self.api_token}
+            api_url = f"https://api-inference.huggingface.co/pipeline/feature-extraction/{self.pretrained_model_name_or_path}"
```
JoanFM, line 150: this `api_url` can be set in the `constructor`. Not need to instatiate at every call N times

JoanFM: so does the header

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/__init__.py
maximilianwerk: I think it is nicer inside this function, but outside the `while` loop. Almost no additional computation, but the API call code is put into just one place.

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -120,9 +139,27 @@ def amp_accelerate(self):
         else:
             return nullcontext()
 
-    def _compute_embedding_from_model_output(self, outputs: 'BaseModelOutput', input_tokens: Dict):
+    def _api_call(self, query):
```
JoanFM, line 142: what is the return type? How from `json` is it so easy to convert to `Torch.Tensor`?

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
```diff
@@ -224,3 +226,24 @@ def test_no_cls_token(test_metas, params):
...
+def test_encoding_results_api(test_metas, pooling_strategy):
+    encoder = TransformerTorchEncoder(
+        metas=test_metas,
+        pretrained_model_name_or_path='sentence-transformers/distilbert-base-nli-stsb-mean-tokens',
+        pooling_strategy=pooling_strategy,
+        api_token='fake_token',
+    )
+
+    api_mock = MagicMock(return_value=np.random.rand(2, 9, 768).tolist())
+    encoder._api_call = api_mock
```
JoanFM, line 241: the _api_call return `json` does not return a `dict`?

JoanFM: Wouldn't it be better to mock the `external request` call better than the complete `_api_call` so that that is the code tested. Also we can assert the parameters sent to the API. To make sure what we send to HuggingFace?

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/__init__.py
maximilianwerk: Because it is already a list of a list of floats. Just the right format.

encoders/nlp/TransformerTorchEncoder/tests/test_transformertorchencoder.py
maximilianwerk: done

encoders/nlp/TransformerTorchEncoder/__init__.py
```diff
@@ -188,10 +224,20 @@ def encode(self, data: 'np.ndarray', *args, **kwargs) -> 'np.ndarray':
...
+                        **input_tokens
+                    )
+                    if isinstance(outputs, torch.Tensor):
+                        return outputs.cpu().numpy()
+                    hidden_states = outputs.hidden_states
 
             if not isinstance(outputs, torch.Tensor):
-                return self._compute_embedding_from_model_output(outputs, input_tokens)
+                return self._compute_embedding(hidden_states, input_tokens)
             else:
```
JoanFM, line 242: I think this `if` else is not anymore needed.

JoanFM (changes requested):

encoders/nlp/TransformerTorchEncoder/__init__.py
maximilianwerk: Good catch. Thanks!

JoanFM (approved):

PR #7461, opened 2021-04-11
feat: add truncated SVD encoder

Harshdeep1996:
This pull request adds a [Truncated Singular Value Decomposition (SVD)](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html) encoder from the Scikit-Learn library.

The truncated SVD dimensionality reduction algorithm does not center the data before the computation of SVD (and hence does not constrain the memory for large datasets) which makes it efficient for working with sparse (input) matrices.

### DIFF
--- a/encoders/numeric/TruncatedSVDEncoder/Dockerfile
+++ b/encoders/numeric/TruncatedSVDEncoder/Dockerfile
@@ -1,4 +1,4 @@
-FROM jinaai/jina
+FROM jinaai/jina as base
 
 # setup the workspace
 COPY . /workspace
@@ -8,6 +8,8 @@ WORKDIR /workspace
 RUN pip install -r requirements.txt
 
 # for testing the image
+FROM base
 RUN pip install pytest && pytest
 
+FROM base
 ENTRYPOINT ["jina", "pod", "--uses", "config.yml"]
--- a/encoders/numeric/TruncatedSVDEncoder/Dockerfile
+++ b/encoders/numeric/TruncatedSVDEncoder/Dockerfile
@@ -1,0 +1,13 @@
+FROM jinaai/jina
+
+# setup the workspace
+COPY . /workspace
+WORKDIR /workspace
+
+# install the third-party requirements
+RUN pip install -r requirements.txt
+
+# for testing the image
+RUN pip install pytest && pytest
+
+ENTRYPOINT ["jina", "pod", "--uses", "config.yml"]
--- a/encoders/numeric/TruncatedSVDEncoder/README.md
+++ b/encoders/numeric/TruncatedSVDEncoder/README.md
@@ -1,0 +1,13 @@
+# TruncatedSVDEncoder
+
+`TruncatedSVDEncoder` encodes data from an `ndarray` of size `BatchSize x Dimension` into an `ndarray` of size `BatchSize x EmbeddingDimension` using [Truncated Singular Value Decomposition (SVD)](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html).  Truncated SVD is a dimension reduction technique which does not center the data before computation which makes it efficient for sparse matrices.
+
+## Usage:
+
+Initialise the encoder using the parameters mentioned below:
+
+| `param name`    | `param_remarks`                              |
+| --------------- | ---------------------------------------------|
+| `output_dim`    | dimensionality of output embedding           |
+| `algorithm`     | algorithm to be used to run SVD              |
+| `max_iter`      | number of iterations for the algorithm to run|
--- a/encoders/numeric/TruncatedSVDEncoder/README.md
+++ b/encoders/numeric/TruncatedSVDEncoder/README.md
@@ -1,13 +1,16 @@
 # TruncatedSVDEncoder
 
-`TruncatedSVDEncoder` encodes data from an `ndarray` of size `BatchSize x Dimension` into an `ndarray` of size `BatchSize x EmbeddingDimension` using [Truncated Singular Value Decomposition (SVD)](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html).  Truncated SVD is a dimension reduction technique which does not center the data before computation which makes it efficient for sparse matrices.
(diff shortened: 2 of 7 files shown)


encoders/numeric/TruncatedSVDEncoder/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM jinaai/jina:1.1.2
```
JoanFM, line 1: Please, remove the version tag of jina. This is updated by dependabot when building new versions

encoders/numeric/TruncatedSVDEncoder/README.md
```diff
@@ -0,0 +1,13 @@
+# TruncatedSVDEncoder
+
+`TruncatedSVDEncoder` encodes data from an `ndarray` of size `BatchSize x Dimension` into an `ndarray` of size `BatchSize x EmbeddingDimension` using [Truncated Singular Value Decomposition (SVD)](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html).  Truncated SVD is a dimension reduction technique which does not center the data before computation which makes it efficient for sparse matrices.
```
JoanFM, line 3: It is not clear, does it work with sparse input or sparse output? In any case, it is not clear in documentation or method signatures

JoanFM (changes requested):

encoders/numeric/TruncatedSVDEncoder/README.md
Harshdeep1996: From the documentation and the examples, it works and can operate on sparse inputs. Should I add that in the documentation/signatures?

encoders/numeric/TruncatedSVDEncoder/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM jinaai/jina
```
nan-wang, line 1:
```suggestion
FROM jinaai/jina as base
```

encoders/numeric/TruncatedSVDEncoder/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# setup the workspace
+COPY . /workspace
+WORKDIR /workspace
+
+# install the third-party requirements
+RUN pip install -r requirements.txt
+
+# for testing the image
+RUN pip install pytest && pytest
```
nan-wang, line 11:
```suggestion
FROM base
RUN pip install pytest && pytest
```

encoders/numeric/TruncatedSVDEncoder/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENTRYPOINT ["jina", "pod", "--uses", "config.yml"]
```
nan-wang, line 13:
```suggestion
FROM base
ENTRYPOINT ["jina", "pod", "--uses", "config.yml"]
```

encoders/numeric/TruncatedSVDEncoder/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        max_iter: int = 200,
+        *args,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self.output_dim = output_dim
+        self.max_iter = max_iter
+        self.algorithm = algorithm
+        self.is_trained = False
+        self.model = None
```
nan-wang, line 41: It is better to declare the `model` in the `post_init` just in case that the `model` can not be properly pickled.
```suggestion
```

encoders/numeric/TruncatedSVDEncoder/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        """Load TruncatedSVD model"""
+        super().post_init()
+        if not self.model:
+            from sklearn.decomposition import TruncatedSVD
+
+            self.model = TruncatedSVD(
+                n_components=self.output_dim,
+                algorithm=self.algorithm,
+                n_iter=self.max_iter,
+            )
```
nan-wang, line 53:
```suggestion
            )
    
    @require_train 
    @batching
    def encode(self, data, *args, **kwargs):
        super().encode(data, *args, **kwargs)
        
```

nan-wang (changes requested):

encoders/numeric/TruncatedSVDEncoder/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    More details can be found
+    `here <https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html>`_
+
+    .. note::
+        :class:`TruncatedSVDEncoder` must be trained before calling ``encode()``.
+    """
+
+    def __init__(
+        self,
+        output_dim: int = None,
```
JoanFM, line 32: We should have a better default value, None I guess will not be supported by the backend

encoders/numeric/TruncatedSVDEncoder/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        algorithm: str = "randomized",
```
JoanFM, line 33: can you please use single quote?

encoders/numeric/TruncatedSVDEncoder/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        max_iter: int = 200,
+        *args,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self.output_dim = output_dim
+        self.max_iter = max_iter
+        self.algorithm = algorithm
+        self.is_trained = False
```
JoanFM, line 42: Remove is trained and require train. This may be deprecated soon

JoanFM (changes requested):
This model should be able to load any pretrained model and not just declare a model to be trained.

encoders/numeric/TruncatedSVDEncoder/__init__.py
Harshdeep1996: Good catch, sure, I will update that, thanks.

rutujasurve94:
@Harshdeep1996 could you make changes as per review so we can merge this PR? thanks for your contribution.

nan-wang:
This issue is stale and closed because it has been open for more than 20 days with no activity.

PR #7558, opened 2021-04-12
fix: documentation URL for ImageOnnxEncoder

Roshanjossey:


theUnkownName (approved):

Yongxuanzhang (approved):
Clearly, we need to add docs for
https://github.com/jina-ai/jina-hub/blob/master/encoders/image/ImageOnnxEncoder/README.md
But we can do this in another pr.

PR #7564, opened 2021-04-13
feat add schema hosting

Kelton8Z:


.github/workflows/schema.yml
```diff
@@ -0,0 +1,32 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: copy-schema-to-api-repo
+        run: |
+          python3 ../../scripts/pydanticJSONgenerator.py
+          mv ../../schema.json ../../api/schemas/schema.json
```
deepankarm, line 29: - `../../api/schemas/schema.json` Please note that the local path for cloning is `schema` not `api`.
- Can you rename the final file to something like `hub_executor_schema.json`, `schema.json` is too generic?

.github/workflows/schema.yml
```diff
@@ -0,0 +1,32 @@
...
+          submodules: true
+      - uses: actions/checkout@v2
+        with:
+          repository: jina-ai/api
+          path: schema
+          token: ${{ secrets.JINA_DEV_BOT }}
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - uses: actions/setup-node@v2
```
deepankarm, line 23: You can remove this. We don't need node here.

.github/workflows/schema.yml
deepankarm, line 29: `mv ../../schema.json` Is this the correct src?

.github/workflows/schema.yml
```diff
@@ -0,0 +1,32 @@
+name: Update JSON schema 
+
+on:
+  push:
+		paths:
+		- ../../schemas/'schema.json'
```
deepankarm, line 6: Should this be `schema.json`? or `scripts/pydanticJSONgenerator.py`?

deepankarm: You can add a `workflow_dispatch` here for testing manually.

scripts/pydanticJSONgenerator.py
```diff
@@ -0,0 +1,61 @@
...
...lear': ' BSD 3-clause Clear license', 'cc': ' Creative Commons license family', 'cc0-1.0': ' Creative Commons Zero v1.0 Universal', 'cc-by-4.0': ' Creative Commons Attribution 4.0', 'cc-by-sa-4.0': ' Creative Commons Attribution Share Alike 4.0', 'wtfpl': ' Do What The F*ck You Want To Public License', 'ecl-2.0': ' Educational Community License v2.0', 'epl-1.0': ' Eclipse Public License 1.0', 'eupl-1.1': ' European Union Public License 1.1', 'agpl-3.0': ' GNU Affero General Public License v3.0', 'gpl': ' GNU General Public License family', 'gpl-2.0': ' GNU General Public License v2.0', 'gpl-3.0': ' GNU General Public License v3.0', 'lgpl': ' GNU Lesser General Public License family', 'lgpl-2.1': ' GNU Lesser General Public License v2.1', 'lgpl-3.0': ' GNU Lesser General Public License v3.0', 'isc': ' ISC', 'lppl-1.3c': ' LaTeX Project Public License v1.3c', 'ms-pl': ' Microsoft Public License', 'mit': ' MIT', 'mpl-2.0': ' Mozilla Public License 2.0', 'osl-3.0': ' Open Software License 3.0', 'postgresql': ' PostgreSQL License', 'ofl-1.1': ' SIL Open Font License 1.1', 'ncsa': ' University of Illinois/NCSA Open Source License', 'unlicense': ' The Unlicense', 'zlib': ' zLib License'})
```
deepankarm, line 21: I guess, we can change to this the way other Enums are defined.

deepankarm (changes requested):

.github/workflows/schema.yml
```diff
@@ -0,0 +1,37 @@
+name: Update JSON schema
+
+on: #workflow_dispatch
+  push
+
+    #paths:
+    #- '../../schemas/schema.json'
```
nan-wang, line 7: Update the schema under `jina-ai/api` only when `schema.json` is changed.
```suggestion
on:
  push
    branches:
    - master
    paths:
    - '../../schemas/schema.json'
```

nan-wang (changes requested):

.github/workflows/schema.yml
Kelton8Z: Done

.github/workflows/schema.yml
Kelton8Z: Sure

.github/workflows/schema.yml
Kelton8Z: using https://github.com/marketplace/actions/push-a-file-to-another-repository instead of mv now

.github/workflows/schema.yml
Kelton8Z: Changed to scripts/pydanticJSONgenerator.py

deepankarm (approved):

PR #7565, opened 2021-04-13
feat: enable changing channel axis

Yongxuanzhang:
I wonder if we could have similar code as 
https://github.com/jina-ai/jina-hub/blob/26faf53637a66d3a0ea30c1b8875e3276faafe4c/encoders/image/ImageTorchEncoder/__init__.py#L88
This is useful in some cases

### DIFF
--- a/encoders/image/CLIPImageEncoder/__init__.py
+++ b/encoders/image/CLIPImageEncoder/__init__.py
@@ -19,10 +19,15 @@ class CLIPImageEncoder(BaseTorchEncoder):
     :param kwargs: Additional positional arguments.
     """
 
-    def __init__(self, model_name: str = 'ViT-B/32',
+    def __init__(self,
+                 model_name: str = 'ViT-B/32',
+                 channel_axis: int = 1,
                  *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.model_name = model_name
+        self.channel_axis = channel_axis
+        # axis 0 is the batch
+        self._default_channel_axis = 1
 
     def post_init(self):
         """Load a model from clip specified in `model_name`."""
@@ -45,6 +50,9 @@ class CLIPImageEncoder(BaseTorchEncoder):
         :param kwargs: Additional positional arguments.
         :return: A `BatchSize x EmbeddingDimension` numpy array.
         """
+
+        if self.channel_axis != self._default_channel_axis:
+            data = np.moveaxis(data, self.channel_axis, self._default_channel_axis)
         import torch
 
         input_torchtensor = torch.from_numpy(data.astype('float32'))
--- a/encoders/image/CLIPImageEncoder/manifest.yml
+++ b/encoders/image/CLIPImageEncoder/manifest.yml
@@ -8,7 +8,7 @@ description: |
 url: https://jina.ai
 vendor: Jina AI Limited
 documentation: https://github.com/jina-ai/jina-hub/blob/master/encoders/image/CLIPImageEncoder/README.md
-version: 0.0.2
+version: 0.0.3
 license: apache-2.0
 keywords: [Cross-Modal, image, CLIP, computer vision]
 type: pod
--- a/encoders/image/CLIPImageEncoder/tests/test_clipimageencoder.py
+++ b/encoders/image/CLIPImageEncoder/tests/test_clipimageencoder.py
@@ -48,7 +48,8 @@ def test_clip_image_encoder():
     np.testing.assert_almost_equal(embeddeding_np, expected, decimal=3)
 
 
-def test_clip_image_encoder_batch():
+@pytest.mark.parametrize('channel_axis', [-1, 1])
+def test_clip_image_encoder_batch(channel_axis):
 
     # Input
     man_piercing_image_path = os.path.join(cur_dir, '../imgs/man_piercing.jpg')
@@ -57,9 +58,10 @@ def test_clip_image_encoder_batch():
     im_tensor_clip_input = preprocess(im).unsqueeze(0)
     im_tensor_clip_np = im_tensor_clip_input.detach().numpy()
     batch = np.vstack([im_tensor_clip_input, im_tensor_clip_input, im_tensor_clip_input])
-
+    if channel_axis == -1:
+        batch = np.moveaxis(batch, 1, -1)
     # Encoder embedding 
-    encoder = CLIPImageEncoder()
+    encoder = CLIPImageEncoder(channel_axis=channel_axis)
     print_array_info(batch, 'batch')
     embeddeding_batch_np = encoder.encode(batch)
     print_array_info(embeddeding_batch_np, 'embeddeding_batch_np')


JoanFM (changes requested):
I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.

Have u checked why it is not working for you adding rhe ImageNormalizer before?

Yongxuanzhang:
> I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.
> 
> Have u checked why it is not working for you adding rhe ImageNormalizer before?

I copy-paste the configuration of ImageNormalizer in our cross model example but still failing. I haven't got too much time to look into it. I guess it's not a big issue but just need some time to check

Yongxuanzhang:
> I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.
> 
> Have u checked why it is not working for you adding rhe ImageNormalizer before?

This decreases the dependency of the clip encoder on ImageNormalizer. We can either configure it at ImageNormalizeror here.
Yet the disadvantage I can see is that this will introduce more parameter, which may make ppl confused.

JoanFM:
> > I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.
> > Have u checked why it is not working for you adding rhe ImageNormalizer before?
> 
> This decreases the dependency of the clip encoder on ImageNormalizer. We can either configure it at ImageNormalizeror or here.
> Yet the disadvantage I can see is that this will introduce more parameter, which may make ppl confused.

The problem I see is that putting this here, may make the channel being changed more than once. After all, ImageNormalization is always needed.  Also it may be confusing because then there are 2 ways of achieving the same thing. 

And since there is no way you will avoid the need to use ImageNormalizer, I would not add this here

Yongxuanzhang:
> > > I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.
> > > Have u checked why it is not working for you adding rhe ImageNormalizer before?
> > 
> > 
> > This decreases the dependency of the clip encoder on ImageNormalizer. We can either configure it at ImageNormalizeror or here.
> > Yet the disadvantage I can see is that this will introduce more parameter, which may make ppl confused.
> 
> The problem I see is that putting this here, may make the channel being changed more than once. After all, ImageNormalization is always needed. Also it may be confusing because then there are 2 ways of achieving the same thing.
> 
> And since there is no way you will avoid the need to use ImageNormalizer, I would not add this here

Yes, I agree. Maybe I need to figure out what's the issue with ImageNormalizer.

Yongxuanzhang:
> > > I am not sure if we should go this way. Since normally ImageNormalizer is needed before most of our ImageEncoders.
> > > Have u checked why it is not working for you adding rhe ImageNormalizer before?
> > 
> > 
> > This decreases the dependency of the clip encoder on ImageNormalizer. We can either configure it at ImageNormalizeror or here.
> > Yet the disadvantage I can see is that this will introduce more parameter, which may make ppl confused.
> 
> The problem I see is that putting this here, may make the channel being changed more than once. After all, ImageNormalization is always needed. Also it may be confusing because then there are 2 ways of achieving the same thing.
> 
> And since there is no way you will avoid the need to use ImageNormalizer, I would not add this here

I doubt this code could work as expected, 
image

JoanFM:
Why is this failing?

```
!ImageNormalizer
with:
  target_size: 224
  img_mean: [0.485, 0.456, 0.406]
  img_std: [0.229, 0.224, 0.225]
  channel_axis: 0
metas:
  name: normalizer
requests:
  on:
    IndexRequest:
      - !CraftDriver
        with:
          executor: normalizer
    SearchRequest:
      - !CraftDriver
        with:
          executor: normalizer
    ControRequest:
      - !ControlReqDriver {}
```

Yongxuanzhang:
> Why is this failing?
> 
> ```
> !ImageNormalizer
> with:
>   target_size: 224
>   img_mean: [0.485, 0.456, 0.406]
>   img_std: [0.229, 0.224, 0.225]
>   channel_axis: 0
> metas:
>   name: normalizer
> requests:
>   on:
>     IndexRequest:
>       - !CraftDriver
>         with:
>           executor: normalizer
>     SearchRequest:
>       - !CraftDriver
>         with:
>           executor: normalizer
>     ControRequest:
>       - !ControlReqDriver {}
> ```
For example, 
because my input data before craft() is (300,300,3), if channel_axis: 0, 
then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.

JoanFM:
> > Why is this failing?
> > ```
> > !ImageNormalizer
> > with:
> >   target_size: 224
> >   img_mean: [0.485, 0.456, 0.406]
> >   img_std: [0.229, 0.224, 0.225]
> >   channel_axis: 0
> > metas:
> >   name: normalizer
> > requests:
> >   on:
> >     IndexRequest:
> >       - !CraftDriver
> >         with:
> >           executor: normalizer
> >     SearchRequest:
> >       - !CraftDriver
> >         with:
> >           executor: normalizer
> >     ControRequest:
> >       - !ControlReqDriver {}
> > ```
> 
> For example,
> because my input data before craft() is (300,300,3), if channel_axis: 0,
> then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.

So put channel_axis to be -1. It is a parameter.

Yongxuanzhang:
> > > Why is this failing?
> > > ```
> > > !ImageNormalizer
> > > with:
> > >   target_size: 224
> > >   img_mean: [0.485, 0.456, 0.406]
> > >   img_std: [0.229, 0.224, 0.225]
> > >   channel_axis: 0
> > > metas:
> > >   name: normalizer
> > > requests:
> > >   on:
> > >     IndexRequest:
> > >       - !CraftDriver
> > >         with:
> > >           executor: normalizer
> > >     SearchRequest:
> > >       - !CraftDriver
> > >         with:
> > >           executor: normalizer
> > >     ControRequest:
> > >       - !ControlReqDriver {}
> > > ```
> > 
> > 
> > For example,
> > because my input data before craft() is (300,300,3), if channel_axis: 0,
> > then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.
> 
> So put channel_axis to be -1. It is a parameter.

No, -1 won't work as well 😂 . Because Clip expects it to be 0. There's no way to make it work if I didn't modify the ImageNormalizer or the clip encoder

JoanFM:
> > > > Why is this failing?
> > > > ```
> > > > !ImageNormalizer
> > > > with:
> > > >   target_size: 224
> > > >   img_mean: [0.485, 0.456, 0.406]
> > > >   img_std: [0.229, 0.224, 0.225]
> > > >   channel_axis: 0
> > > > metas:
> > > >   name: normalizer
> > > > requests:
> > > >   on:
> > > >     IndexRequest:
> > > >       - !CraftDriver
> > > >         with:
> > > >           executor: normalizer
> > > >     SearchRequest:
> > > >       - !CraftDriver
> > > >         with:
> > > >           executor: normalizer
> > > >     ControRequest:
> > > >       - !ControlReqDriver {}
> > > > ```
> > > 
> > > 
> > > For example,
> > > because my input data before craft() is (300,300,3), if channel_axis: 0,
> > > then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.
> > 
> > 
> > So put channel_axis to be -1. It is a parameter.
> 
> No, -1 won't work as well 😂 . Because Clip expects it to be 0. There's no way to make it work if I didn't modify the ImageNormalizer or the clip encoder

Ok, so the shape of ur input image is different as in the crossmodal search. 

I think it makes sense to add target_channel_axis as parameter

Yongxuanzhang:
> > > Why is this failing?
> > > ```
> > > !ImageNormalizer
> > > with:
> > >   target_size: 224
> > >   img_mean: [0.485, 0.456, 0.406]
> > >   img_std: [0.229, 0.224, 0.225]
> > >   channel_axis: 0
> > > metas:
> > >   name: normalizer
> > > requests:
> > >   on:
> > >     IndexRequest:
> > >       - !CraftDriver
> > >         with:
> > >           executor: normalizer
> > >     SearchRequest:
> > >       - !CraftDriver
> > >         with:
> > >           executor: normalizer
> > >     ControRequest:
> > >       - !ControlReqDriver {}
> > > ```
> > 
> > 
> > For example,
> > because my input data before craft() is (300,300,3), if channel_axis: 0,
> > then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.
> 
> So put channel_axis to be -1. It is a parameter.

Just for explanation:
Ideally, I want this:
(300,300,3)->(3,224,244) with crafter->(num_batch, 3,224,224) with clip encoder. 
With current crafter and clip encoder:

- channel_axis=-1 for crafter: (300,300,3)->(224,224,3) with crafter->(num_batch,224,224,3) with clip encoder. the clip encoder will have error
- channel_axis=0 for crafter: (300,300,3)->(224,224,3) with crafter, then there will be errors in crafter.

Yongxuanzhang:
> > > > > Why is this failing?
> > > > > ```
> > > > > !ImageNormalizer
> > > > > with:
> > > > >   target_size: 224
> > > > >   img_mean: [0.485, 0.456, 0.406]
> > > > >   img_std: [0.229, 0.224, 0.225]
> > > > >   channel_axis: 0
> > > > > metas:
> > > > >   name: normalizer
> > > > > requests:
> > > > >   on:
> > > > >     IndexRequest:
> > > > >       - !CraftDriver
> > > > >         with:
> > > > >           executor: normalizer
> > > > >     SearchRequest:
> > > > >       - !CraftDriver
> > > > >         with:
> > > > >           executor: normalizer
> > > > >     ControRequest:
> > > > >       - !ControlReqDriver {}
> > > > > ```
> > > > 
> > > > 
> > > > For example,
> > > > because my input data before craft() is (300,300,3), if channel_axis: 0,
> > > > then `raw_img = _load_image(blob, self.channel_axis)` will fail, because the code expect it to be -1.
> > > 
> > > 
> > > So put channel_axis to be -1. It is a parameter.
> > 
> > 
> > No, -1 won't work as well 😂 . Because Clip expects it to be 0. There's no way to make it work if I didn't modify the ImageNormalizer or the clip encoder
> 
> Ok, so the shape of ur input image is different as in the crossmodal search.
> 
> I think it makes sense to add target_channel_axis as parameter

yes, exactly. I will close this and add that in the ImageNormalizer

JoanFM:
Good, thanks for the explanation!!

PR #7583, opened 2021-04-15
feat: plumber pdf segmenter

Yongxuanzhang:
This PR intends to use a more advanced PDF extractor library - pdf plumber to extract text. This is used for PDF search, and may also add more functions from this library, like position information, etc.

- [x] use [PDFplumber](https://pypi.org/project/pdfplumber/) to extract text as one chunk
- [x] use [pdftitle]( https://pypi.org/project/pdftitle/) to extract title as one chunk


segmenters/nlp/PDFPlumberSegmenter/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+        :param uri: File name of PDF
+        :type uri: str
+        :param buffer: PDF file in bytes
+        :type buffer: bytes
+        :param mime_type: the type of data
+        :returns: A list of documents with the extracted data
+        :rtype: List[Dict]
+        """
+        chunks = []
```
JoanFM, line 35: If we need this, we should add a FliterQL driver before this executor, but this should not be part here in my opinion

segmenters/nlp/PDFPlumberSegmenter/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+So if we have a PDF with 2 images and text, we will have 3 chunks; one per each image and one per the text.
+
+
+## 🏃️ Flow
+First you define your flow here:
+    
+    f = Flow().add(uses='PDFExtractorSegmenter', array_in_pb=True)
+
+If you have a PDF with images and text, the ```array_in_pb=True``` is necessary.
```
JoanFM, line 15: why is this necessary?

segmenters/nlp/PDFPlumberSegmenter/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+And then we search in: 
+
+    with f:
+        f.search(input_fn=search_generator(path=path, buffer=None), on_done=validate_mix_fn)
```
JoanFM, line 20: input_fn is deprecated

JoanFM (dismissed):

segmenters/nlp/PDFPlumberSegmenter/__init__.py
Yongxuanzhang: You mean `chunks = []`? Other segments also have this. Or you mean the mimetype?

segmenters/nlp/PDFPlumberSegmenter/README.md
Yongxuanzhang: I haven't updated the readme. This is a copy-paste from another pdf segmenter. 😄

segmenters/nlp/PDFPlumberSegmenter/__init__.py
Yongxuanzhang: Ah yes, we can add a driver. But this is easier to use. And I plan to add some warning information here to suggest users to use driver to filter only pdf data.

JoanFM: I am not convinced with this. This should not be this executors responsability I think

Yongxuanzhang: How can we solve this situation? I have 3 segments working sequentially, 1 text, 1 image, and 1 pdf. If I use FliterQL at first for the text segmenter, how can I keep the query data of images and pdf?

JoanFM: Why do u have these Sequentially?

Yongxuanzhang: I tried to use them in parallel but will get issues. So I use the compound to stack these 3 together.

JoanFM: ok keep it like this, but is weird

Yongxuanzhang: Yeah, true. I also don't like it. 😂 
I came up with 2 ways for this case:

- Use the 3 segmenters in parallel and each equipped with a filter driver. This is more aligned with our design. But I got very strange issues and cannot reproduce each time. I will try to reproduce it later and maybe open an issue.
- Use the 3 segmenters in sequential, and add this check inside the segmenters.

segmenters/nlp/PDFPlumberSegmenter/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+class PDFPlumberSegmenter(BaseSegmenter):
+    """
+    :class:`PDFPlumberSegmenter` Extracts data (text and images) from PDF files.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @single(slice_nargs=3)
+    def segment(self, uri: str, buffer: bytes, mime_type: str, *args, **kwargs) -> List[Dict]:
```
maximilianwerk, line 21: Too much is happening in this function. Please refactor the text extraction and the image extraction in separate functions.

segmenters/nlp/PDFPlumberSegmenter/tests/test_pdfextractorcrafter.py
```diff
@@ -0,0 +1,96 @@
...
+path_img = os.path.join(cur_dir, 'cats_are_awesome_img.pdf')
+
+with open(path_text, 'rb') as pdf:
+    input_bytes_text = pdf.read()
+
+with open(path_img, 'rb') as pdf:
+    input_bytes_image = open(path_img, 'rb').read()
+
+with open(path_img_text, 'rb') as pdf:
+    input_bytes_images_text = pdf.read()
```
maximilianwerk, line 25: This code should go into a fixture and not on module-level.

maximilianwerk (dismissed):
Content is good. Anyhow, I'd like some code restructuring for better maintainability.

segmenters/nlp/PDFPlumberSegmenter/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+PDFPlumberSegmenter is a segmenter used for extracting images and text as chunks from PDF data. It stores each images and text of each page as chunks separately.
+
+## Usage
+
+We use Pod images in several ways:
+
+1. Run with Docker: `docker run`
+   ```bash
+    docker run jinahub/pod.segmenter.pdfplumbersegmenter:0.0.1-1.0.1 --port-in 55555 --port-out 55556
```
cristianmtr, line 11: jina version is diff from dockerfile

Yongxuanzhang: fixed, thanks!

segmenters/nlp/PDFPlumberSegmenter/tests/test_pdfextractorcrafter.py
Yongxuanzhang: Moved to fixture

segmenters/nlp/PDFPlumberSegmenter/__init__.py
Yongxuanzhang: refactored, thx!

nan-wang (approved):
LGTM👍

PR #7585, opened 2021-04-16
fix: remove duplicated code with core

JoanFM:
**3 things are done in the same PR so that the jina hub build passes**

1. Remove duplicated TransformEncoder (exists in the core)
2. Remove MilvusIndexer (conversation in slack about why) (It can only be tested with a Docker image of them and we cannot support testing against a docker during a docker build)
3. Fix ZarrIndexer's batching #7503


bwanglzu (dismissed):
i see this class has been used a lot in jina hub and also in `test_numeric.py`, why it is a duplicate?

bwanglzu:
image

JoanFM:
See where is it imported from. The same class exists in the core

Yongxuanzhang:
The name of this encoder makes me confused as well. I thought it is a transformer encoder at first.

cristianmtr (approved):

PR #7586, opened 2021-04-16
fix: fix fastica cannot train

JoanFM:


encoders/numeric/FastICAEncoder/__init__.py
```diff
@@ -3,41 +3,12 @@
...
-        """Load FastICA model"""
-        super().post_init()
-        if not self.model:
-            from sklearn.decomposition import FastICA
-            self.model = FastICA(
-                n_components=self.output_dim,
-                whiten=self.whiten,
-                max_iter=self.max_iter)
\ No newline at end of file
+    """
```
Yongxuanzhang, line 14: if we removed all the content, is it just a TransformEncoder?

JoanFM: Yes, almost every numeric encoder fits in this  definition.

Since there is no way to do training live in an executor, it does not make sense to have the models without being fitted

Yongxuanzhang: Oh ok, so the idea is that we don't want to train within the encoder? And we will design training separately?

JoanFM: yes

Yongxuanzhang (approved):

PR #7588, opened 2021-04-16
refactor: refactor no training

JoanFM:


Yongxuanzhang (commented):
It's a little weird that we refactor this, and all numeric encoders will be the same? 😂  I just wonder what is the next plan

JoanFM:
> It's a little weird that we refactor this, and all numeric encoders will be the same?  I just wonder what is the next plan

When they encode the data, they do exactly the same, therefore there is no difference. And since we are going for a design that explicitly separates `executors` from their `training`, we remove the specifities. We could remove these `hub` executors, but for backward compatibility we can mantain this

Yongxuanzhang:
> > It's a little weird that we refactor this, and all numeric encoders will be the same?  I just wonder what is the next plan
> 
> When they encode the data, they do exactly the same, therefore there is no difference. And since we are going for a design that explicitly separates `executors` from their `training`, we remove the specifities. We could remove these `hub` executors, but for backward compatibility we can mantain this

Got it

Yongxuanzhang:
> > It's a little weird that we refactor this, and all numeric encoders will be the same?  I just wonder what is the next plan
> 
> When they encode the data, they do exactly the same, therefore there is no difference. And since we are going for a design that explicitly separates `executors` from their `training`, we remove the specifities. We could remove these `hub` executors, but for backward compatibility we can mantain this

Why not finish the training first then refactor all of this? DO they block something?

JoanFM:
> > > It's a little weird that we refactor this, and all numeric encoders will be the same?  I just wonder what is the next plan
> > 
> > 
> > When they encode the data, they do exactly the same, therefore there is no difference. And since we are going for a design that explicitly separates `executors` from their `training`, we remove the specifities. We could remove these `hub` executors, but for backward compatibility we can mantain this
> 
> Why not finish the training first then refactor all of this? DO they block something?

I want to remove the confusion about `training`. Because is non-usable code. I think is better not to have noise

Yongxuanzhang (approved):
LGTM. It might be better to keep it open so others will also review this PR.

PR #7590, opened 2021-04-16
fix: fix classifiers and problems with tensorflow

JoanFM:


classifiers/nlp/ZeroShotTFClassifier/Dockerfile
```diff
@@ -1,28 +1,15 @@
-FROM continuumio/miniconda3 as conda
-
-COPY requirements.txt /
-RUN conda update conda -c conda-forge && \
-    pip install -r requirements.txt --no-cache-dir && \
-    conda clean -afy
-
-FROM nvidia/cuda:11.0-base-ubuntu20.04 as base
```
Yongxuanzhang, line 8: Why do we remove these?

JoanFM: It was impossible to make it work with the `tensorflow` conficts.

So if we want to keep the versions rolling we need to remove

Yongxuanzhang: oh ok

Yongxuanzhang (approved):

PR #7789, opened 2021-04-20
fix: check path to vectorizer exists

davidbp:
The encoder could 'silently break' if the path to the trained vectorizer did not exist. This PR fixes this.


JoanFM (changes requested):
remember to increase verison in manifest

JoanFM (approved):

PR #7795, opened 2021-04-21
fix: sorting properly casting strings to floats

davidbp:
Pysparnn does not behave as expected when there distances are strings [as shown here.](https://github.com/facebookresearch/pysparnn/issues/35) This PR fixes the issue casting them into floats and sorting the results.

### DIFF
--- a/indexers/vector/PysparnnIndexer/__init__.py
+++ b/indexers/vector/PysparnnIndexer/__init__.py
@@ -158,7 +158,7 @@ class PysparnnIndexer(BaseVectorIndexer):
         distances = []
         indices = []
         for record in index_distance_pairs:
-            sorted_record = sorted(record, key=lambda x: float(x[0]),reverse=False)
+            sorted_record = sorted(record, key=lambda x: float(x[0]), reverse=False)
             distance_list, index_list = zip(*sorted_record)
             distances.append(distance_list)
             indices.append(index_list)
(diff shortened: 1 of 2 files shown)


indexers/vector/PysparnnIndexer/__init__.py
```diff
@@ -158,6 +158,7 @@ def query(self, vectors: 'SparseMatrixType', top_k: int, *args, **kwargs):
         distances = []
         indices = []
         for record in index_distance_pairs:
+            sorted_record = sorted(record, key=lambda x: float(x[0]),reverse=False)
```
JoanFM, line 161: this sorted record should be used after in the zip

JoanFM (changes requested):

indexers/vector/PysparnnIndexer/__init__.py
davidbp: yes, commited too fast, fixed.

indexers/vector/PysparnnIndexer/__init__.py
```diff
@@ -158,7 +158,8 @@ def query(self, vectors: 'SparseMatrixType', top_k: int, *args, **kwargs):
         distances = []
         indices = []
         for record in index_distance_pairs:
-            distance_list, index_list = zip(*record)
+            sorted_record = sorted(record, key=lambda x: float(x[0]),reverse=False)
```
JoanFM, line 161:
```suggestion
            sorted_record = sorted(record, key=lambda x: float(x[0]), reverse=False)
```

JoanFM (changes requested):

JoanFM (approved):

Issue #7796, opened 2021-04-21
Implement ImageNormalizer based on Torch Transforms


JoanFM:
ImageNormalizer executor is hard to make it use batching, because it is based on havinf PIL images for cropping.

May be a good idea to have an ImageNormalizer based on PyTorch transforms that can allow batching.

In this case, batching would not be enforced by the batching decorator, but by the batching parameter passed to the Torch transformer itself.

This can impact very positively the performance of Image Search soultions

bsridatta:
Hello, may I ask why the ImageNormalier crafter has cropping and resize functionality, while there are dedicated crafter for both. Is it not redundant or is it just for ease of use?

JoanFM:
Hey @bsridatta,

I believe it is for ease of use. I find it however weird that it forces to crop the image. I think resizing can be quite expected and common, however `cropping` may feel weird

bsridatta:
Understood, if it's for ease of use, the crop is also required to handle aspect ratios as well (16:9 -> 1:1) as resize maintains the ratio.

So the idea here is to have another TorchImageNormalizer crafter that replaces the crop and resize methods with torchvision.transforms that take numpy as tensors without using PIL right?

JoanFM:
The idea would be this yes, or maybe see if it would make sense to have an `ImageTorchTransformation` executor that can be used to do preprocessing of images using `torchvision.transforms` similar to what  the `AlbumentationsCrafter` does

bsridatta:
Cool, I actually overlooked that there was albumentations crafter. 

I think what you propose is a crafter just like AlbumentationsCrafter but with torchvision that performs a list of transforms. 
- the already existing Albumentations is significantly faster torchvision for most tasks and it also supports other frameworks 
- So ideally this `ImageTorchTransformation` would be better only when doing few transforms which are faster than albumentations.
+ Ya, that would be much better than just having TorchImageNormalizer which does 3 things (crop, resize). 
+ I see that it would helpful if someone wants to work with torchvision purely.

So can I try `ImageTorchTransformation` crafter then?

JoanFM:
Hey @bsridatta ,

U can try this. 

The benefit is also that `TorchTransformations` can work on batches which I am not sure `AlbumentationsCrafters` do. 

And also that `torchvision` can be a more popular framework and people may be more comfortable using it

bsridatta:
Got it, thanks. I shall work on this and maybe also try to do a comparison.

[...]

PR #7988, opened 2021-04-27
feat(crafter): add torchvision transformations

bsridatta:
- PR for issue #7796 

- Adding `ImageTorchTransformation` crafter that applies a sequence of torchvision transforms on a batch of images and uses gpu is available.

- This works exclusively with torch tensors and not PIL. Only 2 transforms work exclusively with PIL, RandomChoice, RandomOrder - both do not apply to Jina usecases and the errors are handled. 

- `AlbumentationsCrafter` tries to remove randomness as much as possilbe by overriding parameters with 'always_apply=True'. I think its best to default it to being deterministic as much as we can (if there is a way) and if the user wants randomness he should be able to have it?
 For now this crafter just overiding parameters to have probability as 1 if possible to match `AlbumentationsCrafter`.


crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+                    " check if all kwargs are valid and if transform is valid for tensors."
+                )
+                raise ValueError
+
+            transforms_list.append(transform_instance)
+        self.transforms = T.Compose(transforms_list)
+
+    def craft(
+        self, blob: Union[List["np.ndarray"], "np.ndarray"], *args, **kwargs
+    ) -> List[Dict]:
```
JoanFM, line 113: blob is np.array, not need to handle lost

JoanFM (commented):
I like the idea. Some comments

crafters/image/ImageTorchTransformation/__init__.py
JoanFM: ah no, my bad

JoanFM (changes requested):
Can you please use single quote? as per style?

crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+    :param transforms: List of transforms that are applied sequentially
+    :param args:  Additional positional arguments
+    :param kwargs: Additional keyword arguments
+    """
+
+    def __init__(
+        self,
+        transforms: List[Union[str, Dict[str, Dict]]],
```
JoanFM, line 59: can we have a default value where a Normalization and Resize is applied?
The typical Resize and normalization required for the main Convolutional Neural Network models?

crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            try:
+                tr_class = getattr(T, tr_name)
+            except AttributeError:
+                self.logger.error(
+                    f"The transform class `{tr_name}` does not seem to exist,"
+                    "check that you have not made a typo."
+                )
+                raise ValueError
+
+            if "p" in tr_class.__init__.__code__.co_varnames:
```
JoanFM, line 95: isn't there a better way to do this? Why is this needed?

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: Ya, since it accepts batch and does not have @single

crafters/image/ImageTorchTransformation/tests/test_imagetorchtransformation.py
```diff
@@ -0,0 +1,217 @@
...
+        ImageTorchTransformation([{"RandomVerticalFlip": dict(Fake=1.0)}])
+    # allowed single image array (H, B, C), list of image arrays, image batch array (B, H, W, C)
+    with pytest.raises(AssertionError):
+        ImageTorchTransformation(["RandomVerticalFlip"]).craft(np.zeros((100, 100)))
+    # transform not supported for tensor
+    with pytest.raises(ValueError):
+        ImageTorchTransformation(["RandomChoice"])
+
+
+def test_example():
```
JoanFM, line 170: I think this test would be good to be tested in batches as well

crafters/image/ImageTorchTransformation/__init__.py
JoanFM: yes, actually it will only be List of numpy arrays.

But I am not sure u can stack them

JoanFM:
Are there tests with inputs in the batches with different shapes?

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: > Can you please use single quote? as per style?

I saw most code was single quotes but I also saw the that the style is Black, which enforces double quotes everywhere right? 

So should I change everything to single except for long strings?

JoanFM: change it everywhere, even for long strings

bsridatta: > can we have a default value where a Normalization and Resize is applied?
> The typical Resize and normalization required for the main Convolutional Neural Network models?

Sure, I shall try to match it to ResNet dimensions and Normalize with ImageNet stats, as its mostly used pre-trained nets

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: I did not find any other way to access that variable even in stacko. 
It is required to remove unwanted randomness to the processing. 
Most transforms in torchvison are 'random' i.e they take a prob, usually defaulted to 0.5 with which it is either applied or skipped. Since our usecase is inference we change it to 1.0 so it always applied. Just like in `AlbumentationsCrafter` but it has an easy `always_apply` param!

JoanFM: I would simply update with p=1 as an extra argument. 

Most of the objects will just ignore except those that use it

JoanFM: so just remove the if

crafters/image/ImageTorchTransformation/tests/test_imagetorchtransformation.py
bsridatta: You mean, `test_incorrect_transforms()`? This is only for testing the initialization of `Compose` alone.
`test_allowed_argument_types` Line 76. tests single and batch inputs for the transforms. By batch, I mean a list of (2) images or numpy array (B, H, W, C) 

I guess I didn't get your suggestion.

bsridatta: > Are there tests with inputs in the batches with different shapes?

Yes I have this 
```
@pytest.mark.parametrize(
    "transforms, inputs",
    [
        (["ColorJitter"], org_img()),
        (["ColorJitter"], np.stack([org_img()] * 3)),
        ([{"GaussianBlur": dict(kernel_size=(3, 3))}], [org_img(), org_img()]),
    ],
)
def test_allowed_argument_types(transforms, inputs):
    # All valid formats of transforms and inputs
```

JoanFM: what I mean is that u test the results match for an image given as a single input, not in batches

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: Makes sense. But for reason, stacking was an important test in albumentationsCrafter

JoanFM: but I am wondering now if this stack will work when images of different shapes are passed.

Thats why I insist on testing with a batch of images of different dimensions

bsridatta: Ah! I understand what you mean now. All the 'batch' test that I did was with the same image (shape) even though there were 2-5.

What you suggest is [small_img, big_img] etc. That's a good point, didn't think in that direction.
For any dataset, the images would ideally be of the same dimensions. But definitely, there could be someone who pools images expecting to handle them here with transforms. 
In that scenario, you cant batch anyway, not with this implementation but in general, as it the tensor should be BCHW

One should do use ImageCropper first to make them the same shape and then use this crafter to do other transforms faster in batch.

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: The stage at which I made `p`=1 is before initializing the transform and having an invalid kwarg throws error, tried it and is also testcase. Instead what I did now is to inject into the transform variables with dictionary `update`. This worked and looks neat, I guess. If `p` is used its 1 now, or its just an unwanted value it holds.

bsridatta:
Summing new commits - 
- I have changed to single quotes everywhere. 
- And also added default transform to match ResNet-like architectures with a testcase.
- Made overriding transform prob. a little better

- I don't think it's possible to apply transform on a batch of images of different shapes

crafters/image/ImageTorchTransformation/__init__.py
JoanFM: Let's put that clearly in the `README` that input needs to be of the same shape

bsridatta: Sure, I added a clear note now.

crafters/image/ImageTorchTransformation/requirements.txt
```diff
@@ -1 +1,2 @@
 torchvision==0.9.1
+opencv_python==4.5.1
```
bsridatta, line 2: I am not sure if it's ok to have this in the requirements as it is only used in the test and not anywhere in the crafter. There was a build error stating cv2 was not found. However, I do not see other crafters having OpenCV in requirements. Let me know.

JoanFM:
recheckcla

JoanFM (changes requested):

crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+                self.logger.error(
+                    f'Error instantiating torchvision transforms class `{tr_name}`,'
+                    ' check if all kwargs are valid and if transform is valid for tensors.'
+                )
+                raise ValueError
+
+            # Removes randomness in transforms
+            # does not handle inherently random transform like `RandomErasing`
+            # where p is probabilty of being random
+            transform_instance.__dict__.update({'p': 1.0})
```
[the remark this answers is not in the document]
bsridatta: > The stage at which I made `p`=1 was before initializing the transform and having an invalid kwarg throws error, tried it and is also testcase. Instead what I did now is to inject into the transform variables with dictionary `update`. This worked and looks neat, I guess. If `p` is used its 1 now, or its just an unwanted value it holds.

That is what I meant here. So if you want it in `tr_kwargs` then I have check if the class actually takes `p` as argument with the not so pretty code -`if "p" in tr_class.__init__.__code__.co_varnames:`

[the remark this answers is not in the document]
JoanFM: why is it necessary? normally the inits accept extra kwargs that are just ignored, have u even tried?

crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -0,0 +1,168 @@
+from typing import Dict, List, Union
+
+import numpy as np
+import torch
```
JoanFM, line 4: can you please hide this `import` under:

```
if False:
    import torch
```

just to offer access for type checking, and do the imports only where they are needed? This way, we do not pollute the `jina check` command line

JoanFM (changes requested):

crafters/image/ImageTorchTransformation/__init__.py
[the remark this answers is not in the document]
bsridatta: It is necessary because by default most transforms have `p` set to 0.5 by default which means it may not always be applied. We make sure it is deterministic (across runs) by setting it to 1.

Yes, I sure tried again after you said to yesterday. That is also why the initialization is under try-catch (even in albumentations) and have also included test cases for wrong kwargs as well. 

Below are couple of screenshots showing it 

image
image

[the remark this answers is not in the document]
JoanFM: ok, thanks for the explanation

crafters/image/ImageTorchTransformation/__init__.py
bsridatta: You are right forgot about that, thanks

crafters/image/ImageTorchTransformation/__init__.py
```diff
@@ -1,10 +1,11 @@
 from typing import Dict, List, Union
 
-import numpy as np
-import torch
-
 from jina.executors.crafters import BaseCrafter
 
+if False:
+    import torch
+    import numpy as np
```
JoanFM, line 7: numpy dependency can be on top, because it is a core jina dependency

crafters/image/ImageTorchTransformation/__init__.py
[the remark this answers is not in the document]
bsridatta: I understand your pov. It has something to do with the transform class implementation not having **kwargs as a parameter.

bsridatta:
@JoanFM Are there any more changes to be made? I have referred to a discussion in `FaceNetSegmenter` regarding dependencies for opencv-python and changed the docker file to match it. I hope it fixes the errors from the last build.

JoanFM:
recheckcla

JoanFM:
> @JoanFM Are there any more changes to be made? I have referred to a discussion in `FaceNetSegmenter` regarding dependencies for opencv-python and changed the docker file to match it. I hope it fixes the errors from the last build.

Everything is good, just need you to sign the CLA.

bsridatta:

JoanFM (approved):

JoanFM:
Thank you very much for this awesome contribution!

bsridatta:
Feels great, thanks for the help!

PR #8081, opened 2021-04-29
feat: query on batches leveldbindexer

davidbp:
This PR should add query working on batches.

### DIFF
--- a/indexers/keyvalue/LevelDBIndexer/__init__.py
+++ b/indexers/keyvalue/LevelDBIndexer/__init__.py
@@ -43,6 +43,8 @@ class LevelDBIndexer(BinaryPbIndexer):
             aux = self.query_handler.get(bytes(key))
             if aux is not None:
                 vs.append(Parse(aux.decode('utf8'), Document()))
+            else:
+                vs.append(None)
 
         return vs
 
--- a/indexers/keyvalue/LevelDBIndexer/__init__.py
+++ b/indexers/keyvalue/LevelDBIndexer/__init__.py
@@ -38,7 +38,12 @@ class LevelDBIndexer(BinaryPbIndexer):
         """
         from google.protobuf.json_format import Parse
 
-        vs = [Parse(self.query_handler.get(bytes(key)).decode('utf8'), Document()) for key in keys]
+        vs = []
+        for key in keys:
+            aux = self.query_handler.get(bytes(key))
+            if aux is not None:
+                vs.append(Parse(aux.decode('utf8'), Document()))
+
         return vs
 
     def add(self, keys: Iterable[str], values: Iterable[bytes], *args, **kwargs) -> None:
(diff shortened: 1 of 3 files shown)


indexers/keyvalue/LevelDBIndexer/tests/test_leveldbindexer.py
```diff
@@ -64,8 +64,9 @@ def apply_actions(save_abspath, index_abspath, actions):
 
 def validate_positive_results(keys, documents, searcher):
     assert len(list(zip(keys, documents))) > 0
+
```
JoanFM, line 67: I'd rather have `result_docs = search.query(keys)` and iterate with zip with `query_docs` and `result_docs`

indexers/keyvalue/LevelDBIndexer/tests/test_leveldbindexer.py
```diff
@@ -74,7 +75,7 @@ def validate_positive_results(keys, documents, searcher):
 def validate_negative_results(keys, searcher):
     assert len(keys) > 0
     for key in keys:
-        result_doc = searcher.query(key)
+        result_doc = searcher.query([key])
```
JoanFM, line 78: same here

JoanFM (changes requested):

JoanFM (commented):
Remember to bump the `manifest` version to trigger CI

indexers/keyvalue/LevelDBIndexer/tests/test_leveldbindexer.py
davidbp: Done

indexers/keyvalue/LevelDBIndexer/__init__.py
```diff
@@ -38,7 +38,12 @@ def query(self, keys: Iterable[str], *args, **kwargs) -> Optional[bytes]:
         """
         from google.protobuf.json_format import Parse
 
-        vs = [Parse(self.query_handler.get(bytes(key)).decode('utf8'), Document()) for key in keys]
+        vs = []
+        for key in keys:
+            aux = self.query_handler.get(bytes(key))
+            if aux is not None:
+                vs.append(Parse(aux.decode('utf8'), Document()))
```
JoanFM, line 45: you need to return `None` in the `list` if the `key` does not `exist`.

Otherwise the client will not be able to know which result correspond to which key.

JoanFM (changes requested):

indexers/keyvalue/LevelDBIndexer/__init__.py
JoanFM: if this test passes, this means the tests are wrong. We need to add a test where we query `present` and `unexisting` keys in the same batch

JoanFM: Please fix this and make sure there is a test sensitive to this

JoanFM (changes requested):

indexers/keyvalue/LevelDBIndexer/__init__.py
davidbp: changed validate_negative_results to take into account the number of elements in the list (not just that missing keys should get None values, but that the number of missing keys should be equal to the number of None values)

JoanFM (approved):

PR #8083, opened 2021-04-30
feat: add batching query capability to mongodbindexer

davidbp:
This PR allows query to work on `keys:Iterable[str]`

### DIFF
--- a/indexers/keyvalue/MongoDBIndexer/__init__.py
+++ b/indexers/keyvalue/MongoDBIndexer/__init__.py
@@ -94,8 +94,7 @@ class MongoDBIndexer(BinaryPbIndexer):
             for key in keys:
                 result.append(mongo_handler.query(key))
 
-        if result:
-            return result
+        return result
 
     def update(self, keys: Iterable[str], values: Iterable[bytes], *args, **kwargs) -> None:
         """Update the document at the given key in the database.
--- a/indexers/keyvalue/MongoDBIndexer/__init__.py
+++ b/indexers/keyvalue/MongoDBIndexer/__init__.py
@@ -83,14 +83,16 @@ class MongoDBIndexer(BinaryPbIndexer):
         """Get the handler to MongoDB."""
         return self.get_query_handler()
 
-    def query(self, key: str, *args, **kwargs) -> Optional[bytes]:
+    def query(self, keys: Iterable[str], *args, **kwargs) -> Optional[bytes]:
         """Query the serialized documents by document id.
 
         :param key: document id
         :return: serialized document
         """
+        result = []
         with self.query_handler as mongo_handler:
-            result = mongo_handler.query(key)
+            for key in keys:
+                result.append(mongo_handler.query(key))
 
         if result:
             return result
--- a/indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
+++ b/indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
@@ -43,7 +43,8 @@ def test_mongodbindexer():
         mongo_indexer.add(keys=keys, values=values)
 
     with MongoDBIndexer() as mongo_query:
-        result = mongo_query.query(key=query_key)
+        results = mongo_query.query(keys=[query_key])
+        result = results[0]
         assert result['_id'] == query_key
         d = Document()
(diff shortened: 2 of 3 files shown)


indexers/keyvalue/MongoDBIndexer/__init__.py
```diff
@@ -83,14 +83,16 @@ def query_handler(self) -> 'MongoDBHandler':
...
+    def query(self, keys: Iterable[str], *args, **kwargs) -> Optional[bytes]:
         """Query the serialized documents by document id.
 
         :param key: document id
         :return: serialized document
         """
+        result = []
         with self.query_handler as mongo_handler:
-            result = mongo_handler.query(key)
+            for key in keys:
```
JoanFM, line 94: would be good to see if the mongo handler can handle itself multiple queries at the same time

davidbp: tried but did not work

JoanFM: see if can be changed to make it work

nan-wang:
```
result = mongo_handler.query(keys)
```

nan-wang: change [mongodbhandler.py:L79](https://github.com/jina-ai/jina-hub/blob/master/indexers/keyvalue/MongoDBIndexer/mongodbhandler.py#L79) to 
```
cursor = self.collection.find({'_id': {"$in": keys}})
```

reference:

https://docs.mongodb.com/manual/tutorial/query-documents/#specify-conditions-using-query-operators

nan-wang: I think this should work.

nan-wang (changes requested):

indexers/keyvalue/MongoDBIndexer/__init__.py
JoanFM: The problem may come from the fact that you cannot keep track of the `failing` keys. 

So maybe it is better to keep the single batch inside the `mongohandler`

davidbp: This change that nan proposes works, the issue is that then I need to iterate over the returned list and verify the ids to check for missing ids;

I can do 
```aux = mongo_handler.query(['fccf9b08-acaf-11eb-bd09-0242ac110002', '221-fakeid-221',  'fccfda14-acaf-11eb-bd09-0242ac110002'])```

But then `len(aux)` is 2, not 3, and I need a loop over items.

File: indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py [hunk 2]
```diff
@@ -73,5 +75,5 @@ def test_mongodbindexer():
 
     with MongoDBIndexer() as mongo_query:
         for key in keys:
```
JoanFM, line 77: can u do `assert mongo_query.query(keys) == [None, None...]`

JoanFM (changes requested):

indexers/keyvalue/MongoDBIndexer/__init__.py
```diff
@@ -83,14 +83,16 @@ def query_handler(self) -> 'MongoDBHandler':
...
+                result.append(mongo_handler.query(key))
 
         if result:
```
JoanFM, line 97:
```suggestion
        return result
```

JoanFM: no need for this check

indexers/keyvalue/MongoDBIndexer/tests/test_mongodbindexer.py
JoanFM: change the way you query in this test

JoanFM (changes requested):

indexers/keyvalue/MongoDBIndexer/__init__.py
davidbp: True,

JoanFM (approved):

